PR #48, opened 2020-03-05
Fix #47 - Add stash support

### LINKED ISSUE #47: Export a "stash" of changes from the previous CRLite run
`create_filter_cascade` should find the differences in the current dataset versus the prior dataset, then store those in a data-stash of deltas.

The delta need only include additions to the revoked or not-revoked sets, not removals -- removals are either set-changes (which would be in the relevant set), or they are expiration events, which won't be checked against the CRLite filter anyway.

The data format should be as compressed as reasonable.

jcjones:
This by default generates an `mlbf` file named `filter.stash`, which contains
all the additions to the filter, collated by issuer and status change.

The file is made of per-issuer blocks, where each block's bytes are:

```
  bytes 0-3: N, number of revoked serials as an unsigned long
  bytes 4-7: M, number of nonrevoked serials as an unsigned long
  bytes 8-9: L, length of the issuer field as a unsigned short
  bytes 10+: hash of issuer subject public key info of length L
  ... then N serials_structs
  ... followed by M serials_structs
```

The serials_struct blocks are defined:
```
  bytes 0-1: S, length of serial field as an unsigned short
  bytes 2+: ASN.1 DER-encoded serial number of length S
```

When reaching EOF, there are no more issuer blocks.

Consumers can collect these `filter.stash` files, load them in chronological
order, and apply them before querying the originating MLBF filter.

### DIFF
--- a/create_filter_cascade/certs_to_crlite.py
+++ b/create_filter_cascade/certs_to_crlite.py
@@ -127,6 +127,7 @@ def genCertLists(args, stats, *, revoked_certs_by_issuer, nonrevoked_certs_by_is
             issuer = os.path.splitext(filename)[0]
             if issuer in args.excludeIssuer:
                 continue
+            issuer_bytes = bytes(issuer, encoding="utf-8")
 
             initIssuerStats(stats, issuer)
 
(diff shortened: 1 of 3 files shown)


jcjones:
Good catch, I'll adjust the documentation and the commit message.

create_filter_cascade/certs_to_crlite.py
```diff
@@ -350,6 +394,22 @@ def saveMLBF(args, stats, cascade):
     sw.end('save')
 
 
+def find_additions(*, old_by_issuer, new_by_issuer):
+    added = {}
+    for issuer_b64 in new_by_issuer:
```
mozkeeler: I wonder if we want to verify that anything that was in the old set but isn't in the new set has expired?

jcjones: I like the idea, but this code doesn't have the knowledge to check that. It could emit all the "losses" that are presumably to expirations which could then be checked after the fact -- what do you think?

mozkeeler: That seems reasonable.

Issue #58, opened 2020-04-06
Add a pod/cronjob for verifying and signing-off Kinto updates


jcjones:

jcjones:
Will be calling https://github.com/mozilla/moz_crlite_query, and requires https://github.com/mozilla/moz_crlite_query/issues/2

PR #63, opened 2020-04-07
Add CRLite query utility

jcjones:
Still Work-In-Progress:

- [ ] Gracefully handle certs with unknown issuers
- [ ] Add `--check-valid` and `--check-revoked` methods to programmatically return error codes if the supplied PEMs don't meet that condition, for scripting
- [ ] Print the cert's subject maybe, but pyasn1 doesn't like pretty-printing RDNs. Consider it anyway.

### DIFF
--- a/crlite_query/crlite_query.py
+++ b/crlite_query/crlite_query.py
@@ -1,0 +1,445 @@
+import argparse
+import collections
+import crlite
+import hashlib
+import logging
+import base64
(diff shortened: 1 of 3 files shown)


jcjones:
This is part of the work for #58.

jcjones:
Moved to https://github.com/mozilla/moz_crlite_query

Issue #67, opened 2020-04-21
Add a Kinto schema entry for timestamp and use it


jcjones:
make it easy for Javascript and pass a JS epoch millisecond timestamp

jcjones:
Basically, for now keep the semantics of the current naming fields, but clients going forward should only obey the timestamp, incremental, and previousID fields.

jcjones:
Proposed new schema for Kinto:

jcjones:
Proposed new schema for Kinto:

```json
{
  "type": "object",
  "required": [
    "incremental",
    "effectiveTimestamp"
  ],
  "properties": {
    "details": {
      "type": "object",
      "title": "Details",
      "required": [
        "name"
      ],
      "properties": {
        "name": {
          "type": "string",
          "title": "Name",
          "description": "Deprecated",
        }
      }
    },
    "incremental": {
      "type": "boolean",
      "title": "Incremental",
      "description": "Is this an incremental update (false if a new cascade, true for updates)."
    },
    "parent": {
      "type": "string",
      "title": "Parent record ID",
      "description": "If this is an incremental update, the ID of the record being updated."
    },
    "effectiveTimestamp": {
      "type": "number",
      "title": "Record effective time",
      "description": "Timestamp of the point of time of this record's dataset, in milliseconds since the epoch."
    }
  },
  "description": "Certificate revocation data."
}
```

mozkeeler:
This sounds good to me.

PR #69, opened 2020-05-15
Prepare to automate the publisher (New schema for Issue #67)

jcjones:
This reworks the publisher tool to link stashes and filters, use the new schema, and obtain its data from Google Cloud Storage rather than local disk, allowing it to run in its own pod.

This latter point means this no longer belongs in `crlite-fullrun`.


moz_kinto_publisher/test_publisher.py
```diff
@@ -0,0 +1,228 @@
...
+        "attachment": {
+            "hash": "hash",
+            "size": 0,
+            "filename": f"{run_id}",
+            "location": "abc",
+            "mimetype": "application/octet-stream",
+        },
+        "incremental": diff,
+        "id": random_id,
+        "last_modified": 1,
```
mozkeeler, line 27: Should this include a parent field if this is a diff?

jcjones: Absolutely should. Thanks!

PR #71, opened 2020-05-15
Fix #68: Don't be fooled by new downloads of out-of-date data.

### LINKED ISSUE #68: aggregate-crls shouldn't be fooled if served old files
Sometimes a CDN might return a CRL from the past. As an example, the run from midnight GMT on 14 May, the CRL http://crl.godaddy.com/gdig2s1-1812.crl as fetched had the following timestamps:
```
        Last Update: May 13 17:48:14 2020 GMT
        Next Update: May 20 17:48:14 2020 GMT
```
Fetched 6 hours later at 0600 GMT on 14 May, the same file’s timestamps were:
```
        Last Update: May 12 17:41:25 2020 GMT
        Next Update: May 19 17:41:25 2020 GMT
```

Right now the CRL downloader makes sure the CRL processes safely before accepting it as “newer” than the one already on disk:

https://github.com/mozilla/crlite/blob/5a9671e9ba4d5a6179094c33e5a6046cb76942c1/go/cmd/aggregate-crls/aggregate-crls.go#L147-L152

The CRL downloader needs to compare those update dates before accepting a new CRL, too.

jcjones:
If the newly downloaded CRL is actually older than the previous CRL, keep
the previous.

Adds more tests to aggregate-crls.

### DIFF
--- a/go/cmd/aggregate-crls/aggregate-crls_test.go
+++ b/go/cmd/aggregate-crls/aggregate-crls_test.go
@@ -47,7 +47,7 @@ func Test_makeFilenameFromUrl(t *testing.T) {
 
 func makeCA(t *testing.T) (*x509.Certificate, interface{}) {
 	caTemplate := &x509.Certificate{
-		SerialNumber: big.NewInt(1),
+		SerialNumber: big.NewInt(time.Now().Unix()),
 		Subject: pkix.Name{
 			CommonName: "Honest Achmed's Used Certificates and CRLs",
 		},
--- a/go/cmd/aggregate-crls/aggregate-crls_test.go
+++ b/go/cmd/aggregate-crls/aggregate-crls_test.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 	"net/url"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


go/cmd/aggregate-crls/aggregate-crls_test.go
```diff
@@ -32,5 +43,166 @@ func Test_makeFilenameFromUrl(t *testing.T) {
 		"http://crl.repository.com/",
 		"http://crl.repository.com/crl"}
 	checkCollision(t, crls2, names)
+}
+
+func makeCA(t *testing.T) (*x509.Certificate, interface{}) {
+	caTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(1),
```
mozkeeler, line 50: With a fixed serial, this will create a lot of certificates with the same issuer/serial but different contents. I guess it's not a big deal if these aren't persistent and don't interact with NSS ever.

jcjones: I mean, might as well use the current epoch timestamp 👍 =D

jcjones:
Thanks for the review!

PR #72, opened 2020-05-22
Don't bypass CRLs when there's an issue with the new download

jcjones:
`aggregate_crls` had code to download a CRL to a temporary location, verify it, and then copy it over the previous CRL if it looked good. However there was a `continue` in there that made the routine throw away the CRL's contents.

This reworks that part of the code to be more testable, and add tests!

Most notably, even if `crlFetchWorkerProcessOne` returns an error, it still might return a valid CRL on disk, which should then be used for processing.

### DIFF
--- a/go/cmd/aggregate-crls/aggregate-crls.go
+++ b/go/cmd/aggregate-crls/aggregate-crls.go
@@ -131,6 +131,10 @@ func (ae *AggregateEngine) crlFetchWorkerProcessOne(ctx context.Context, crlUrl 
 	if err != nil {
 		glog.Warningf("[%s] Could not download %s to %s: %s", issuer.ID(), crlUrl.String(),
 			tmpPath, err)
+		err = os.Remove(tmpPath)
+		if err != nil {
+			glog.Warningf("[%s] Failed to remove invalid tmp file %s: %s", issuer.ID(), tmpPath, err)
+		}
 	} else {
 		// Validate the file and move it to the finalPath
 		cert, err := ae.issuers.GetCertificateForIssuer(issuer)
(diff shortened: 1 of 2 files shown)


go/cmd/aggregate-crls/aggregate-crls.go
```diff
@@ -116,6 +116,63 @@ func (ae *AggregateEngine) findCrlWorker(ctx context.Context, wg *sync.WaitGroup
...
+
+	err = downloader.DownloadFileSync(ctx, ae.display, crlUrl, tmpPath, 3)
+	if err != nil {
+		glog.Warningf("[%s] Could not download %s to %s: %s", issuer.ID(), crlUrl.String(),
+			tmpPath, err)
+	} else {
+		// Validate the file and move it to the finalPath
+		cert, err := ae.issuers.GetCertificateForIssuer(issuer)
+		if err != nil {
+			glog.Fatalf("[%s] Could not find certificate for issuer: %s", issuer.ID(), err)
```
mozkeeler, line 138: Does `glog.Fatalf` terminate the program?

go/cmd/aggregate-crls/aggregate-crls.go
```diff
@@ -116,6 +116,63 @@ func (ae *AggregateEngine) findCrlWorker(ctx context.Context, wg *sync.WaitGroup
...
+	}
+
+	filename := makeFilenameFromUrl(crlUrl)
+	tmpPath := filepath.Join(*crlpath, issuer.ID(), filename+".tmp")
+	finalPath := filepath.Join(*crlpath, issuer.ID(), filename)
+
+	err = downloader.DownloadFileSync(ctx, ae.display, crlUrl, tmpPath, 3)
+	if err != nil {
+		glog.Warningf("[%s] Could not download %s to %s: %s", issuer.ID(), crlUrl.String(),
+			tmpPath, err)
```
mozkeeler, line 133: Is it guaranteed that there is no file at `tmpPath` here? (the other branches clean up the temporary file, so I was wondering if we wanted to do that here as well)

go/cmd/aggregate-crls/aggregate-crls.go
jcjones: It does, which will abort the pod and send alerts that the job couldn't complete.

This would be a pretty fatal error, indicating some failure in `issuers.go` to maintain state.

go/cmd/aggregate-crls/aggregate-crls.go
jcjones: Good call. Cleaning it up.

PR #77, opened 2020-06-10
Fix #45: Do not use Google Firestore

### LINKED ISSUE #45: Remove dependencies on Firestore
The whole Firestore side of CRLite’s backend could be scrapped if we have good backups of the Redis, and the `ct-fetch` instance were changed to store state into Redis as well.

The downside is the loss of the bulk PEM data for future reprocessing purposes, if we needed to scan for other data.

Obviously, could also be changed to BigTable, but the data sizes do make such a thing somewhat costly.

jcjones:
This review set seeks to remove all but transitive dependencies on Google Firestore. https://github.com/jcjones/ct-mapreduce/issues/39 will ultimately remove all further dependencies.


jcjones:
Thanks for the review!

PR #94, opened 2020-06-26
Coordinate the Publisher with Upload Artifacts to Storage script

### LINKED ISSUE #75: Kinto Publisher and Uploader need coordination (race)
There's a potential race where an uploader could be in-progress on the `mlbf` directory when the Publisher reads it, and miss either the `filter` or `filter.stash` files because they haven't been uploaded yet.

The uploader should post a sentinel that indicates its completed that the publisher looks for before attempting to proceed with the chosen ID.

jcjones:
The `workflow/2-upload_artifacts_to_storage` script now uploads a file
`{run_folder}/completed` at completion time, which `moz_kinto_publisher`
watches for (for 5 minutes) before proceeding with uploading CRLite.

I didn't apply this logic to the intermediates because I don't see a similar
chicken-and-egg issue there. It'd fail gracefully.

Fixes #75

In staging e485e61

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -9,6 +9,7 @@ import re
 import requests
 import settings
 import tempfile
+import time
 import workflow
 
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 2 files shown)


jcjones:
> Assuming the sentinel naming is consistent (that is, `run_id` == `Path(idPath.name)`), LGTM.

I'll normalize those two by changing the var name in `2-upload_artifacts_to_storage`. Thanks!

PR #95, opened 2020-06-26
Use ct-mapreduce v1.0.10

jcjones:
This adds the -nobars option and a higher-performance health check

In staging e485e61


jcjones:
Actually using 07d911d

jcjones:
Thanks!

PR #96, opened 2020-06-30
Move all scripts into the base container image.

smarnach:
It turned out that having four different container images to deal with causes more issues with our tooling than I anticipated. As a preliminary way of dealing with this, I propose this PR, which is completely untested at this point. It moves all scripts into the base image, so the actual content of all image tags is the same, and they only differ in things like the command line and the environment variables, which can be changed when running the container. This would allow us to only use the base image on our end.

This PR is a minimal first step towards #92.


jcjones:
After I get some testing of the other PRs, I'll push this one to the dev environment as well.

containers/Dockerfile
```diff
@@ -37,9 +37,10 @@ COPY moz_crlite_lib /app/moz_crlite_lib
 COPY create_filter_cascade /app/create_filter_cascade
 COPY moz_kinto_publisher /app/moz_kinto_publisher
 COPY workflow /app/workflow
+COPY containers/scripts /app/scripts
 COPY setup.py /app/
 RUN pip3 install /app/moz_crlite_lib/ /app/
-
+RUN pip3 install --user moz_crlite_query
```
smarnach, line 43: We may need to move this line after the `USER` directive.

jcjones: I think it's better if it not be modifiable by the user account, so removing `--user` and replacing the relevant path in the calling script with `/usr/local/bin`

jcjones: (That's in b94c10a738707f42af405354172e645a757a813c)

PR #100, opened 2020-07-01
Don't require credentials for ct-fetch

### LINKED ISSUE #97: Don't require credentials for `crlite-fetch`
The fetch tooling should not require credentials. Check and remove the requirement from `crlite-fetch.sh` if so.

jcjones:
Fix #97 by not requiring any GCP service credentials for `ct-fetch`, since without Firestore none are needed.


jcjones:
Works in staging.

jcjones:
All nominal overnight.

PR #114, opened 2020-07-14
 Split "Failed Verify" into more useful audit entries

### LINKED ISSUE #113: "Failed Verify" as a CRL audit term is not explanatory
Most of the "Failed Verify" entries are actually more like "Obtained Old CRL", so we should special-case that event and leave "Failed Verify" to actual signature errors.

jcjones:
Fixes #113. Most of the "Failed Verify" entries are actually more like
 "Obtained Old CRL", so this special-cases those out into "Older Than Previous"
 and also covers Expiration more clearly, since those were not tracked before.

### DIFF
--- a/go/cmd/aggregate-crls/crl-auditor.go
+++ b/go/cmd/aggregate-crls/crl-auditor.go
@@ -101,7 +101,7 @@ func (auditor *CrlAuditor) FailedVerifyUrl(issuer storage.Issuer, crlUrl *url.UR
 	})
 }
 
-func (auditor *CrlAuditor) FailedOlderthanPrevious(issuer storage.Issuer, crlUrl *url.URL, dlAuditor *DownloadAuditor, previous time.Time, this time.Time) {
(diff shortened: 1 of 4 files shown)


jcjones:
In `dev` as 3cd1842d0dc81519db4c564d66c1b46d5aad578c

jcjones:
The "Older Than Previous" entries don't provide DNS information, which is unfortunate. Should fix that.

jcjones:
Hmmm, that _should_ be working, from what I see in the code.

go/cmd/aggregate-crls/crl-auditor.go
```diff
@@ -98,6 +101,23 @@ func (auditor *CrlAuditor) FailedVerifyUrl(issuer storage.Issuer, crlUrl *url.UR
 	})
 }
 
+func (auditor *CrlAuditor) FailedOlderthanPrevious(issuer storage.Issuer, crlUrl *url.URL, dlAuditor *DownloadAuditor, previous time.Time, this time.Time) {
```
kjacobs-moz, line 104: s/Olderthan/OlderThan 
(same with the test function).

jcjones: Thanks!

PR #115, opened 2020-07-14
Catch exceptions that might be thrown by `is_expired`

### LINKED ISSUE #107: Don't crash if the logging-only expiration check fails in moz_kinto_publisher
These lines can raise Exception:
https://github.com/mozilla/crlite/blob/3006b8b4ab08cfc2282fa286b7e172913c3c5b27/moz_kinto_publisher/main.py#L651-L653

On large sets, this could happen for a variety of reasons... but since `expired` is only ever used for audit purposes, we could just emit a warning log instead of crashing.

jcjones:
Fixes #107

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -649,8 +649,11 @@ def publish_intermediates(*, args, ro_client, rw_client):
 
     expired = set()
     for i in to_delete:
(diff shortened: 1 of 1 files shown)


jcjones:
In `dev` as fca87e3031144f3b87cd4abe5f3aaad2f2d43331

PR #117, opened 2020-07-14
Emit stash size information into stats.json

### LINKED ISSUE #109: Emit stash size information into stats.json
Stats.json provides useful information about the MLBF, but not about the stash.

jcjones:
Fixes #109 by logging the filesize of the stash and the count of participating revokers into the `stats.json`


jcjones:
In `dev` as 6c162ba.

Issue #123, opened 2020-07-18
Unable to load the Mozilla issuers: Not a valid PEM


jcjones:
Generation currently broken due to some change at the CCADB

```json
{
  "insertId": "m9idpjiz2k540wg30",
  "jsonPayload": {
    "pid": "17",
    "source": "aggregate-crls.go:576",
    "message": "Unable to load the Mozilla issuers: Not a valid PEM at line 29995"
  },
  "timestamp": "2020-07-18T12:03:06.315637Z",
  "severity": "CRITICAL",
  "logName": "projects/crlite-beta/logs/stderr",
  "sourceLocation": {
    "file": "aggregate-crls.go",
    "line": "576"
  },
  "receiveTimestamp": "2020-07-18T12:03:25.356111905Z"
}
```

Relevant code: https://github.com/mozilla/crlite/blob/3c9708b7e8cca59cb7916f76ce688f52402af1ff/go/rootprogram/issuers.go#L219-L227

jcjones:
There are two lines in the CCADB report that have no PEM, where column 4 is `""`.

```csv
"CN=NII Open Domain CA - G5; O=National Institute of Informatics; C=JP","OU=Security Communication RootCA2; O=SECOM Trust Systems CO.,LTD.; C=JP","DACAA09FFAD8F66EDB6ED69707A827C9E44C1F97C48F0179485F206D707B3E2B",""
"CN=NII Open Domain CA - G5; O=National Institute of Informatics; C=JP","OU=Security Communication RootCA2; O=SECOM Trust Systems CO.,LTD.; C=JP","09983A2632C033EB9449C0EBF21F9B8D4A9975C06FB8E20F4A5042E9E4CC6411",""
```

jcjones:
The best approach for this is likely to be like what we do for CRLs -- store the last "good" in the persistent folder, and fall-back to that if the freshly-downloaded data is bad.

I'll abstract the CRL code that handles that.

PR #125, opened 2020-07-21
Persist CCADB downloads if and only if they load OK

jcjones:
This is a fairly large patch, but broken into logical commits. It:
* Refactors the CRL Audit tools into the `downloader` class
* Constructs a "Verifying Downloader" based on the `aggregate-crls` approach of download-and-verify-and-rename used for CRLs
* Uses the new verifying downloader for CRLs and adapts tests
* Uses the new verifying downloader for the Mozilla Issuers data from the CCADB
* Adapts the `aggregate-crls` tool and its script to store the CCADB into the persistent folder

Local log with a valid file on-disk and the currently-problematic one from #123 served by the CCADB:

```
I0720 23:31:50.984102   64274 config.go:176] Loaded config file from /Users/jcjones/.ct-fetch.ini
I0720 23:31:50.996565   64274 aggregate-crls.go:554] Progress bar refresh rate is every 125ms.
I0720 23:31:50.996601   64274 engine.go:76] aggregate-crls is starting. Local statistics will emit every: 1m0s
W0720 23:31:54.285309   64274 issuers.go:79] Failed verify of https://ccadb-public.secure.force.com/mozilla/MozillaIntermediateCertsCSVReport: Not a valid PEM at line 29950
W0720 23:31:54.387808   64274 issuers.go:111] Error encountered loading CCADB data, but able to proceed with previous data. Error: Not a valid PEM at line 29950
```

### DIFF
--- a/go/downloader/verifying-downloader.go
+++ b/go/downloader/verifying-downloader.go
@@ -32,38 +32,40 @@ func DownloadAndVerifyFileSync(ctx context.Context, verifyFunc DownloadVerifier,
 		}
 	}()
 
+	attemptFallbackToExistingFile := func(err error) (bool, error) {
+		existingValidErr := verifyFunc.IsValid(finalPath)
+		if existingValidErr == nil {
+			// The existing file at finalPath is OK.
+			return true, err
+		}
+		// We don't log to the auditor here since the local file being bad isn't necessarily this run's fault,
+		// and it will be handled later in aggregate-crls if it is relevant at that stage.
+		combinedError := fmt.Errorf("[%s] Couldn't verify already-on-disk path %s. Local error=%s, Caused by=%s",
+			identifier.ID(), finalPath, existingValidErr, err)
+		glog.Error(combinedError)
+		return false, combinedError
+	}
+
 	dlErr := DownloadFileSync(auditCtx, display, crlUrl, tmpPath, maxRetries)
 	if dlErr != nil {
 		auditor.FailedDownload(identifier, &crlUrl, dlTracer, dlErr)
 		glog.Warningf("[%s] Failed to download from %s to tmp file %s: %s", identifier.ID(), crlUrl.String(), tmpPath, dlErr)
 
-		existingValidErr := verifyFunc.IsValid(finalPath)
-		if existingValidErr == nil {
-			// The existing file at finalPath is OK.
-			return true, dlErr
-		}
-		return false, dlErr
+		return attemptFallbackToExistingFile(dlErr)
 	}
 
 	dlValidErr := verifyFunc.IsValid(tmpPath)
 	if dlValidErr != nil {
 		auditor.FailedVerifyUrl(identifier, &crlUrl, dlTracer, dlValidErr)
 
-		existingValidErr := verifyFunc.IsValid(finalPath)
-		if existingValidErr == nil {
-			// The existing file at finalPath is OK.
-			return true, dlValidErr
-		}
-
-		auditor.FailedVerifyPath(identifier, finalPath, existingValidErr)
-		glog.Errorf("[%s] Couldn't verify already-on-disk path %s: %s", identifier.ID(), finalPath, existingValidErr)
-		return false, existingValidErr
+		return attemptFallbackToExistingFile(dlValidErr)
 	}
 
 	renameErr := os.Rename(tmpPath, finalPath)
 	if renameErr != nil {
 		glog.Errorf("[%s] Couldn't rename %s to %s: %s", identifier.ID(), tmpPath, finalPath, renameErr)
-		return false, renameErr
+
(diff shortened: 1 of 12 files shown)


go/downloader/verifying-downloader.go
```diff
@@ -0,0 +1,73 @@
...
+	dlErr := DownloadFileSync(auditCtx, display, crlUrl, tmpPath, maxRetries)
+	if dlErr != nil {
+		auditor.FailedDownload(issuer, &crlUrl, dlTracer, dlErr)
+		glog.Warningf("[%s] Failed to download from %s to tmp file %s: %s", issuer.ID(), crlUrl.String(), tmpPath, dlErr)
+
+		existingValidErr := verifyFunc.IsValid(finalPath)
+		if existingValidErr == nil {
+			// The existing file at finalPath is OK.
+			return true, dlErr
+		}
```
smarnach, line 44: Should we call `auditor.FailedVerifyPath(issuer, finalPath, existingValidErr)` here?

go/downloader/verifying-downloader.go
```diff
@@ -0,0 +1,73 @@
...
+
+		existingValidErr := verifyFunc.IsValid(finalPath)
+		if existingValidErr == nil {
+			// The existing file at finalPath is OK.
+			return true, dlValidErr
+		}
+
+		auditor.FailedVerifyPath(issuer, finalPath, existingValidErr)
+		glog.Errorf("[%s] Couldn't verify already-on-disk path %s: %s", issuer.ID(), finalPath, existingValidErr)
+		return false, existingValidErr
```
smarnach, line 60: It's unclear whether we want to return `dlValidErr` or `existingValidErr`. If we reach this line, both of them are non-nil. In line 47, you opted for `dlError`, and here for the error with the existing file.

go/downloader/verifying-downloader.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	renameErr := os.Rename(tmpPath, finalPath)
+	if renameErr != nil {
+		glog.Errorf("[%s] Couldn't rename %s to %s: %s", issuer.ID(), tmpPath, finalPath, renameErr)
+		return false, renameErr
```
smarnach, line 66: It's unclear whether we should unconditionally return `false` when renaming fails. We could also fall back to validating the existing file, as we do when downloading or validating the download fails.

The code for the existing file could be made less repetitive by moving it to the end of the function and using `goto`. This may be one of the cases where `goto` actually _improves_ readability. :)

go/rootprogram/issuers.go
```diff
@@ -46,18 +49,69 @@ type EnrolledIssuer struct {
...
-	return mi.downloadAndParse(kMozCCADBReport)
+	ctx := context.Background()
+
+	display := mpb.New(
+		mpb.WithOutput(ioutil.Discard),
+	)
+
+	dataUrl, err := url.Parse(mi.reportUrl)
+	if err != nil {
+		glog.Fatalf("Couldn't parse CCADB URL of %s: %s", mi.reportUrl, err)
```
smarnach: Do we want to `return err` here?

jcjones: I probably should, for clarity. Fatal will terminate as-is, it's possible `go vet` will complain with the unreachable code, but I'll handle it. 

A further question is whether it should Fatal, but IMO since this is a configuration error, it should indeed.

jcjones: For clarity, sure. `glog.Fatal` aborts the program, but `go vet` doesn't complain about unreachable code, so SGTM!

(As I mentioned in the other comment, I think this does deserve an abort for the URL being misconfigured.)

go/downloader/verifying-downloader.go
jcjones: Yes, though I'm refactoring to avoid messing that up :) Good catch

go/downloader/verifying-downloader.go
jcjones: I'm making a synthetic combined error so we don't lose either one. Good catch.

go/downloader/verifying-downloader.go
jcjones: I've reworked this to use a closure. That's much cleaner - thanks!

jcjones:
> I've got a few questions about these changes – some of them about the control flow, and some probably because my knowledge of Go is rather superficial.
> 
> (Sorry for the delay – I've been on PTO for the last two days.)

Thanks so much for the detailed review! Updates coming...

go/rootprogram/issuers.go
smarnach: I even looked at the documentation, but apparently I skimmed it too quickly and missed that it calls `os.Exit(255)`. In that case I don't mind either way, but having the return statement doesn't hurt.

PR #127, opened 2020-07-24
 Emit a gauge for "age of CCADB data" from aggregate-crls

### LINKED ISSUE #126: Emit a gauge for "age of CCADB data" from aggregate-crls
It would be a useful metric to see if the CCADB data starts aging, e.g., no updates for a long while.

jcjones:
Fixes #126

### DIFF
--- a/go/cmd/aggregate-crls/aggregate-crls.go
+++ b/go/cmd/aggregate-crls/aggregate-crls.go
@@ -19,6 +19,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/armon/go-metrics"
 	"github.com/golang/glog"
 	"github.com/google/certificate-transparency-go/x509"
 	"github.com/google/certificate-transparency-go/x509/pkix"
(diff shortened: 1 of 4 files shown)


go/cmd/aggregate-crls/aggregate-crls.go
```diff
@@ -574,8 +575,11 @@ func main() {
 
 	if err != nil {
 		glog.Fatalf("Unable to load the Mozilla issuers: %s", err)
+		return
 	}
 
+	metrics.SetGauge([]string{"IssuersAge"}, float32(mozIssuers.DatasetAge().Hours()))
```
smarnach, line 581: I think most of our duration metrics are in seconds rather than hours, but I don't really know if there is an actual convention. In any case, it's most important to be consistent within CRLite, so if we already emit other duration metrics in hours, this one should be hours, too.

jcjones: OK, seconds sounds good to me. In theory this should always be very low, since it should be fresh at each run. But this would be one where if it's more than a cronjob-period old, something has gone wrong.

Will change to seconds!

jcjones:
Rebased.

PR #129, opened 2020-08-06
Rename Docker user to "app" and change UID.

smarnach:
... to comform with our [guidelines][1].

[1]: https://github.com/mozilla-services/dockerflow#dockerfile-requirements

This may make parts of the data on the physical volume inaccessible when deployed to the dev environment. In stage, I will simply delete the physical volume and regenerate it from scratch. If this causes too much trouble, we can discuss alternatives, e.g. adding `chmod -R app:app /persistent` temporarily to fix the data.

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -19,7 +19,7 @@ RUN apt update && apt install -y ca-certificates && \
     apt-get -y clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN adduser --system --home /app appuser
+RUN adduser --system --uid 10001 --group --home /app app
 
 ENV crlite_log /var/log
 ENV crlite_processing /processing
@@ -27,7 +27,7 @@ ENV crlite_persistent /persistent
 ENV crlite_workflow /app/workflow
 ENV crlite_bin /app
 
-RUN mkdir /processing && chown appuser /processing && chmod 777 /processing
+RUN mkdir /processing && chown app /processing && chmod 777 /processing
 VOLUME /var/log /processing /persistent
 
 COPY --from=builder /build/bin /app/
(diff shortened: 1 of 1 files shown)


jcjones:
Pushed to `dev` as fe4e526.

This does require either a modification to these lines to be recursive:
https://github.com/mozilla/crlite/blob/6ba9eb32fa25a510301524b88a39477005b98246/containers/crlite-generate/pod.yaml#L18-L24

or to reinitialize the persistent volume. I'm just reinitializing, it's not a big deal. Has some impacts short-term about bad CRL downloads, but it'll even out quickly enough.

In prod we might want to be more careful and/or backup the persistent volume from time to time though.

jcjones:
The reinitialization of the persistent folder for that permissions update has had no discernible effect on the filters. 20200808-0 was the first run after the change.
image

smarnach:
> In prod we might want to be more careful and/or backup the persistent volume from time to time though.

So far I've assumed we do not need to back the persistent volume up, since it can simply be regenerated if needed. What's the worst case scenario if we don't have backups?

jcjones:
> > In prod we might want to be more careful and/or backup the persistent volume from time to time though.
> 
> So far I've assumed we do not need to back the persistent volume up, since it can simply be regenerated if needed. What's the worst case scenario if we don't have backups?

The worst case scenario is that we lose some revocations (filter gets false negatives) until we get the CRLs all back.

We could run the `aggregate-crls` job multiple times, if we wanted, to try harder to download all the data. We also can (and should) use the CRL audit data to go after the CAs whose data is regularly inconsistent.

Anyway, I think it's OK to assume it's not backed up. This would be a rare event, and much easier to recover from than loss of the Redis database.

smarnach:
Ok, this sounds acceptable to me. If we lose the physical volume, we need some manual intervention anyway, so we can also manually audit the data in that case.

PR #134, opened 2020-09-11
Make the signoff tool use structured logs

jcjones:


jcjones:
In dev as 6a2bb1b

smarnach (commented):
If I read the code correctly, this will print a dictionary in Python's `repr()` format. Can we make this JSON instead? Stackdriver allows to filter by individual fields for JSON log messages, which is sometimes handy.

jcjones:
What is your preference for structured loggers, @smarnach

smarnach:
I would have a slight preference for JSON logs, as stated, since they make log filtering in Stackdriver nicer. We'll probably want to filter the logs whenever some domain is failing – with 10,000+ domains on the list, it won't be obvious on first glance anymore.

smarnach:
Ah, sorry, you mean what structured logging lib? I'd just wrap the dict we have in `json.dumps()` and be done with it.

jcjones:
Thanks, sorry, got pulled away as I was typing. OK, I didn't realize Stackdriver could handle any format like that, and not need some preamble to indicate it was JSON.

jcjones:
Now logs look like:

```
INFO:query_cli:Database was updated at 2020-09-14 09:29:39.953520, skipping.
INFO:query_cli:Status: 2084 Intermediates, Current filter: 2020-09-08T18:08:25+00:00Z-full with 24 layers and 14705968 bit-count, 12 stash files with 33078 stashed revocations, up-to-date as of 2020-09-11 18:08:34+00:00 (2 days, 22:23:48.943301 ago).
INFO:crlite_query:{"name": "getfirefox.com:443", "issuer": "CN=DigiCert SHA2 Secure Server CA,O=DigiCert Inc,C=US", "state": "Valid", "enrolled_in_crlite": true, "cert_id": "CertID(e6426f344330d0a8eb080bbb7976391d976fc824b5dc16c0d15246d5148ff75c-019d2b994ec99445c735d2a6d739e43a)", "via_filter": null, "via_stash": null}
```

smarnach:
I'm actually not sure whether this will work out of the box, but whit standard JSON I should be able to convince fluentd to parse this into structured data that Stackdriver can understand, which would be very difficult with Python `repr()` format.

smarnach (approved):

jcjones:
OK. We can always look at replacing https://pypi.org/project/glog/ with something that supports structured logging out of the box.

PR #137, opened 2020-09-30
Remote --filter-bucket command-line argument.

smarnach:
This argument doesn't exists anymore, since the filter is read from Remote Settings instead.


jcjones (approved):
Oops. :) Thanks!

PR #139, opened 2020-10-01
Remove the read-only Kinto client from the publisher.

smarnach:
The security-state-staging collection is not supposed to be publicly accessible. For some reason we haven't figured out yet, it currently is [publicly accessible in prod](https://bugzilla.mozilla.org/show_bug.cgi?id=1668535), and it used to be publicly accessible in stage until recently. At least for now, the correct way to read that collection is an authenticated read on the Remote Settings writer instance.

CC @sciurus

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -276,9 +276,7 @@ def main():
             )
         )
 
-    log.info(
-        f"Connecting... RW={settings.KINTO_RW_SERVER_URL}"
-    )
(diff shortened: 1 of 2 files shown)


jcjones:
hmm; this is going to change  execution of the tool when the VPN isn't accessible. Maybe that doesn't matter much, but it'll affect `dev` a bit, as right now there's the clean separation that all the `rw_` instances will fail, but methods that contain `ro_` can pass, hence the split.

I'm not sure I want to take this; perhaps we can just set `KINTO_RO_SERVER_URL` to the same URL in the config and document it instead? Or I'm open to other ideas

PR #141, opened 2020-10-13
Docker login before build

jcjones:
Beginning 2020-11-2 Docker Hub is rate limiting anonymous image pulls per IP address. This will impact our jobs that run on CircleCI. So let's login first every time we're talking to Docker. We already do that for the publish task, but we should on the build as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ commands:
         type: string
 
     steps:
+      - docker/check
       - run:
           name: Generate version.json
           command: |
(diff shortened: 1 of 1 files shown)


smarnach (commented):
I _think_ some of the jobs here pull docker images to run the CircleCI jobs on, e.g. `circleci/golang:1.13` and `cimg/base:stable`, so we probably need to add an `auth` step. I'll look into the details tomorrow, since I decided to try and go to bed earlier in the future.

smarnach (approved):
Thanks for the update, this looks good now.

PR #142, opened 2020-10-20
Make the Kinto Publisher check that signatures are needed before requesting signatures

### LINKED ISSUE #140: Signer should only sign records it didn't sign already.
We are currently running the signoff job on an hourly schedule, and it seems to re-signs the preview data every hour. Re-signing does not cause clients to redownload the data, so it probably doesn't matter much, but I somehow feel we should only sign the records once.

jcjones:
As it was, the `crlite_sign` and `intermediates_sign` methods always signed when
called, so every time the k8s job ran to sign, the signer actually did work.
This change is to check the state of the collection to be `"to-review"` before
requesting a signature.

Since I don't have a kinto-signer instance working locally, this was tested on
stage by manipulating the crlite records and rolling it back when done.

Fixes #140

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -156,7 +156,7 @@ class PublisherClient(Client):
                 + f"{response.content.decode('utf-8')}"
             )
 
-    def collection_needs_review(self, *, collection=None):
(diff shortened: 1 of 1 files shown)


smarnach (approved):
Thanks, looks good to me!

PR #146, opened 2020-11-18
Redefine enrollment to be 'all CRLs have to be valid'

### LINKED ISSUE #145: aggregate-crls should require all known CRLs to be valid
aggregate-crls currently considers an issuer enrolled if any CRL known to the Redis DB is valid
https://github.com/mozilla/crlite/blob/a906dcb43eb6fc0d4f5af961696af434c7bda802/go/cmd/aggregate-crls/aggregate-crls.go#L324-L328

Perhaps it should be only if all CRLs known to the Redis DB are valid. That's a high bar, and it makes it effective policy that CRLs must be kept up-to-date for the lifetime of the _issuer_, not just the certs in question, as right now there's no means to track when a CRL DP "drops out" of the active set.

jcjones:
Fixes #145

Require all issuer CRLs to be valid and up-to-date to include that issuer in the enrolled list, instead of at least one CRL be valid.

To make this more easily verifiable, also emit information to `crl-audit.json` for all processed CRLs, and always include the URL at which a given CRL is canonically located, allowing auditing tools to verify the number of CRLs processed for each issuer, and the whole status for each canonical URL.

This will dramatically increase the size of the crl-audit JSON, but then again it was pretty small before. It shouldn't be significant, especially compared to the increased auditability.

### DIFF
--- a/go/cmd/aggregate-crls/aggregate-crls.go
+++ b/go/cmd/aggregate-crls/aggregate-crls.go
@@ -124,7 +124,7 @@ type CrlVerifier struct {
 }
 
 func (cv *CrlVerifier) IsValid(path string) error {
-	_, err := loadAndCheckSignatureOfCRL(path, cv.expectedIssuerCert)
+	_, _, err := loadAndCheckSignatureOfCRL(path, cv.expectedIssuerCert)
 	return err
 }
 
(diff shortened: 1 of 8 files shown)


jcjones:
pushed to `dev` combined with #147 as 23bd9b5.

jcjones:
Pushed https://github.com/mozilla/crlite/pull/146/commits/6e9996789b55d5e56409768202d55cfe829518f9 to dev

kjacobs-moz (approved):

jcjones:
Pushed https://github.com/mozilla/crlite/pull/146/commits/6707d5421bdab35843c465ee49fef783dcd4cc60 to `dev` too

jcjones:
Several days of data looks good. Let's call this a version.

PR #147, opened 2020-11-18
Uplift the useful parts of jcjones/ct-mapreduce into mozilla/crlite

### LINKED ISSUE #119: Move the ct-fetch tool, and the whole of the ct-mapreduce library, into crlite
There's only historical reasons for the [ct-mapreduce library](https://github.com/jcjones/ct-mapreduce) to exist, and it does complicate matters to have to account for its version for all other things.

I should mark that repo deprecated, move the relevant bugs here, and consolidate the code into this repository.

jcjones:
The important parts of https://github.com/jcjones/ct-mapreduce are what CRLite
needs, and the rest of that package is probably broken at best. So let's not
actually depend on it, let's move it all together into one place.

Fixes #119

Also handles https://github.com/jcjones/ct-mapreduce/issues/48.


mozkeeler (approved):
yes, this appears to be golang

jcjones:
pushed to `dev` combined with #146 as 23bd9b5.

jcjones:
No issues in `dev`

PR #155, opened 2021-04-20
incorporate ct-fetch from https://github.com/jcjones/ct-mapreduce

mozkeeler:
This incorporates ct-fetch from https://github.com/jcjones/ct-mapreduce into this repository, which keeps related code together and reduces some extra steps at build time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The keys used into the CRLite data structure consist of the SHA256 digest of the
 It's possible to run the tools locally, though you will need local instances of Redis and Firestore. First, install the tools and their dependnecnies
 
 ```sh
-go install -u github.com/jcjones/ct-mapreduce/cmd/ct-fetch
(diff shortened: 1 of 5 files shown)


smarnach (approved):
Looks good to me.

I'm not sure we need to pull in the whole history of the ct-mapreduce repo just to add that single file – I probably wouldn't, but don't have a strong opinion on it.

mozkeeler:
Thanks! Do you know if there's a way to pull in just the commits relevant to that file? That's what I was trying to do...

smarnach:
You need to rewrite the commits with `git-filter-branch` to achieve that. You may get away with a `--subdirectory-filter`, since it's the only file in the directory, e.g.
```
git filter-branch --subdirectory-filter cmd/ct-fetch -- --all
```
in the ct-mapreduce repo runs fine, and the result looks reasonable at first glance. In general these things can get tricky quickly, e.g. if you want to extract a single file that doesn't happen to be the only file in a directory.

PR #175, opened 2021-12-02
Fix Issue #172: Improvements to moz_kinto_publisher 

### LINKED ISSUE #153: moz_kinto_publisher should publish a new full filter at least every 10 days
An open question we've had is how long between publishing full filters is optimal? We don't want users to download data unnecessarily, but stashes are not efficient, so having too many of them reduces performance. In particular, see [bug 1689796](https://bugzilla.mozilla.org/show_bug.cgi?id=1689796). It turns out that even loading an accumulated stash can cause significant delays in verifying certificates. The approach I'm going to take in that bug is to load the stash on a background task as soon as the cert storage instance is created, and let verification essentially proceed without stashes until that task finishes. This means that for some verifications, we could be missing revocation information made available since the last full filter was downloaded. If we publish a new full filter every 10 days, the window of opportunity for attackers is no broader for crlite than that of OCSP, which is what we use now. Additionally, this should result in smaller accumulated stashes, and thus better performance overall.

### LINKED ISSUE #172: Review and simplify moz_kinto_publisher
The publish job recently got stuck in a bad state after an intermittent failure in the generate task. The generate task recovered on the next run, but it left artifacts from the interrupted job in the crlite-filters directory. The publish job got stuck because it waits for a job completion notice for each collection of artifacts:
https://github.com/mozilla/crlite/blob/3bdafdc6796b020e8983b5f1e4694a32b1d13c72/moz_kinto_publisher/main.py#L1252-L1254

This is one of several issues we've seen with the publish job recently. I'll use this issue to track progress as I review and simplify the publishing logic.

jschanck:
I've tried to simplify the logic in moz_kinto_publisher to make it a bit easier to extend. I've also removed the signoff logic, as it was configured as a no-op in our production environment.

This code has run successfully in the dev environment (to the extent that it can w/o write access to collections).

Also the new sanity checks now flag some errors in currently published data. When this merges to main the sanity checks will cause the bad data to be overwritten, but I've saved copies so that I can look into root causes.

Resolves #153
Resolves #172


moz_kinto_publisher/main.py
```diff
@@ -1178,34 +1178,52 @@ def crlite_verify_run_id_sanity(*, run_db, identifiers_to_check):
...
+    try:
+        crlite_verify_record_sanity(existing_records=existing_records)
+    except SanityException as se:
+        log.error(f"Failed to verify existing record sanity: {se}")
+        return default
+
+    # Get a list of run IDs that are newer than any existing record.
+    # These are candidates for inclusion in an incremental update.
+
+    # A run ID is a "YYYMMDD" date and an index, e.g. "20210101-3".
```
mozkeeler, line 834: Typo: `YYYYMMDD`

moz_kinto_publisher/main.py
```diff
@@ -1178,34 +1178,52 @@ def crlite_verify_run_id_sanity(*, run_db, identifiers_to_check):
...
+        if run_date < cut_date or (run_date == cut_date and run_idx <= cut_idx):
+            old_run_ids.append(run_id)
+        else:
+            new_run_ids.append(run_id)
 
-    if not published_run_ids:
-        return {"clear_all": True, "upload": [run_db.most_recent_id()]}
+    # If we don't have data from old runs, publish a full filter.
+    if not old_run_ids:
+        log.error(f"We do not have data to support existing records.")
```
mozkeeler: Is this really an error? Maybe I'm not understanding...

moz_kinto_publisher/main.py
```diff
@@ -1178,34 +1178,52 @@ def crlite_verify_run_id_sanity(*, run_db, identifiers_to_check):
...
+        return default
 
-    # verify sanity of the run_identifiers
+    # If the new runs fail a sanity check, publish a full filter.
```
mozkeeler, line 859: I prefer `consistency` over words like `sanity` in these situations.

moz_kinto_publisher/main.py
```diff
@@ -1248,71 +1235,46 @@ def publish_crlite(*, args, ro_client, rw_client):
     existing_records = ro_client.get_records(
         collection=settings.KINTO_CRLITE_COLLECTION
     )
+    # Sort existing_records for crlite_verify_record_sanity
```
mozkeeler, line 873: Might be helpful to add something like ", which gets called in `crlite_determine_publish`"

moz_kinto_publisher/main.py
```diff
@@ -254,85 +254,106 @@ def _get_attributes(self):
...
-    crlite_enrolled: bool
-    pemAttachment: AttachedPem
     cert: x509.Certificate
-    certHash: str
-    subjectDN: bytes
+    crlite_enrolled: bool
     derHash: bytes
+    kinto_id: str
+    pemAttachment: AttachedPem
+    pemHash: str
```
mozkeeler, line 178: `pemData` too?

moz_kinto_publisher/main.py
```diff
@@ -254,85 +254,106 @@ def _get_attributes(self):
...
+        self.pemHash = None
+        self.subjectDN = None
 
-        if len(self.pubKeyHash) < 26:
-            raise IntermediateRecordError(f"Invalid intermediate hash: {kwargs}")
+        parseError = IntermediateRecordError(f"Malformed record: {kwargs}")
+
+        # Local records have a "pem" field.
+        # RemoteSettings records have "attachment".
+        # TODO(jms): These should be versioned.
```
mozkeeler, line 196: Let's file an issue and reference it here.

mozkeeler (approved):
This looks good. I basically only had cosmetic comments.

moz_kinto_publisher/main.py
jschanck: Me too, thanks for pointing it out.

moz_kinto_publisher/main.py
jschanck: OK

moz_kinto_publisher/main.py
jschanck: Yep, done.

moz_kinto_publisher/main.py
jschanck: Opened Issue #180 and updated comment in source.

moz_kinto_publisher/main.py
jschanck: I think it's important for debugging and auditing purposes that we keep the data we use to generate a filter for the ~10 days when that filter is in use. We already keep the data, so this consistency test just asks whether we've experienced catastrophic data loss.

PR #177, opened 2021-12-02
Land dev branch backlog

jschanck:
Dana, the first three commits here were on the dev branch when I started working. I've reviewed them and everything looks good except for one [dropped call](https://github.com/mozilla/crlite/commit/2b08211f078cf7ebfb4a9271a7e8e66e53bfcc2d#diff-460fe8ae10fb702dc5989fca37f03c2388180b7c5f77939344d239dc9a0dbe19L185) to IssuerMetadata.Accumulate.

Requesting your review just for the last commit in this sequence, which adds that call back in. The last commit also cleans up some unnecessarily complex (and unsafe) lock manipulations in the Accumulate method. I can break the cleanup into a separate PR if you'd like.

This code has been tested on dev.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ ENV/
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 problems
+
+# vim swap files
+.*.swp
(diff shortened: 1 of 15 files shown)


mozkeeler (approved):
If we interpret this as you've reviewed my changes and I've reviewed your changes, I think we're good here.

PR #179, opened 2021-12-02
Fix Issue #178: Publish ct-logs collection through moz_kinto_publisher

### LINKED ISSUE #178: Publish CT log metadata to new ct-logs Remote Settings collection
The new ct-logs Remote Settings collection will be used by ct-fetch to determine which CT logs to scrape. It will also provide useful metadata to Firefox users.

The moz_kinto_publisher script should be updated so that it publishes an exhaustive list of usable CT logs to the ct-logs collection. For now we can copy Google's list from https://www.gstatic.com/ct/log_list/v3/log_list.json.

Records in the ct-logs collection have a "crlite_enabled" flag, which moz_kinto_publisher should set to False. Logs will be manually enrolled by ct-logs editors.

Blocks #144

jschanck:
This PR is only for commit 3f6b89ba54faff8bdd73245fd002a424da4576f, I've just stacked it on top of #175.

This has been tested on dev.

Depends #175
Resolves #178


moz_kinto_publisher/main.py
```diff
@@ -961,6 +961,136 @@ def publish_crlite(*, args, ro_client, rw_client):
...
+        ctlog for operator in log_list_json["operators"] for ctlog in operator["logs"]
+    ]
+
+    # Translate |upstream_logs_raw| to our schema (and remove unused fields)
+    upstream_logs = [
+        {
+            "crlite_enrolled": False,
+            "description": ctlog["description"],
+            "key": ctlog["key"],
+            "logID": ctlog["log_id"],
```
mozkeeler, line 994: Either I'm confused or `log_id` in the comment about our schema should be `logID`.

moz_kinto_publisher/main.py
```diff
@@ -961,6 +961,136 @@ def publish_crlite(*, args, ro_client, rw_client):
...
+            hashlib.sha256(base64.b64decode(upstream_log["key"])).digest()
+        )
+        if rfc6962_log_id != upstream_log["logID"].encode("utf8"):
+            raise SanityException(
+                f"Google log list contains incorrectly computed logID {upstream_log}"
+            )
+
+    # Fetch our existing Kinto records
+    known_logs = ro_client.get_records(collection=settings.KINTO_CTLOGS_COLLECTION)
+    known_lut = {ctlog["logID"]: ctlog for ctlog in known_logs}
```
mozkeeler, line 1019: What does `lut` stand for?
Also, do we want to consistency-check the preexisting kinto data, just in case we've put something bad in there?

mozkeeler (approved):
Cool - this looks good.

moz_kinto_publisher/main.py
jschanck: "Lookup table". Open to suggestions.

I added a test to ensure that the logID fields are unique. With this change, 1) each unique logID in our list corresponds to a unique entry in Google's list; and 2) the "description", "key", "url", and "mmd" fields (i.e. everything but "crlite_enrolled") match the corresponding entries in Google's list. The only field not covered is "crlite_enrolled", which doesn't need any consistency checks.

moz_kinto_publisher/main.py
jschanck: The comment was wrong. Thanks.

moz_kinto_publisher/main.py
mozkeeler: Ok - looks good to me.

PR #186, opened 2021-12-03
Use python-glog for all logging

### LINKED ISSUE #185: Use python-glog instead of stderr in all python scripts
We should use python-glog instead of printing to stderr in all python scripts. We do this for most scripts already, but a few scripts spam our logs with messages that have no severity level.

jschanck:
Resolves #185


mozkeeler (approved):
Makes sense.

PR #189, opened 2021-12-06
moz_kinto_publisher: Check that cert-revocation records have `attachment` field

jschanck:
Adds a consistency check to ensure that every record in the `cert-revocations` collection has an `attachment` field.

This PR also fixes a bug which caused an attachment field not to be created.


mozkeeler (approved):
Ah.

PR #190, opened 2021-12-09
Part 1 of ct-fetch upgrade

jschanck:
This PR makes two major changs to ct-fetch. First we verify the entries that we download against the internal nodes of the CT log's Merkle tree. This "check-summing" will ultimately let us import entries from a source other than the CT log itself. The second commit changes the order in which we download log entries. Reverse order is preferable if we're going to import a large cache of entries from a third party.

### DIFF
--- a/go/cmd/ct-fetch/ct-fetch.go
+++ b/go/cmd/ct-fetch/ct-fetch.go
@@ -36,7 +36,7 @@ import (
 	"github.com/mozilla/crlite/go/engine"
 	"github.com/mozilla/crlite/go/storage"
 	"github.com/vbauerster/mpb/v5"
-	"github.com/vbauerster/mpb/v5/decor"
+	//	"github.com/vbauerster/mpb/v5/decor"
 )
 
 var (
@@ -257,16 +257,22 @@ type LogSyncEngine struct {
 
 // Operates on a single log
 type LogWorker struct {
-	Bar        *mpb.Bar
-	Database   storage.CertDatabase
-	Client     *client.LogClient
-	LogURL     string
-	STH        *ct.SignedTreeHead
-	LogState   *storage.CertificateLog
-	StartPos   uint64
-	EndPos     uint64
-	SaveTicker *time.Ticker
+	Database  storage.CertDatabase
+	Client    *client.LogClient
+	LogURL    string
+	STH       *ct.SignedTreeHead
+	LogState  *storage.CertificateLog
+	WorkOrder LogWorkerTask
+	JobSize   uint64
 }
+
+type LogWorkerTask int
+const (
+	Init LogWorkerTask = iota // Initialize db with one batch of recent certs
+	Backfill                  // Download old certs
+	Update                    // Download new certs
+	Sleep                     // Wait for an STH update
+)
 
 func NewLogSyncEngine(db storage.CertDatabase) *LogSyncEngine {
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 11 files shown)


go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -278,100 +448,240 @@ func (ld *LogSyncEngine) NewLogWorker(ctLogUrl string) (*LogWorker, error) {
...
+	//
+	oldSize := firstIndex
+	newSize := lastIndex + 1
+	if oldSize == 0 {
+		// Special case: the consistency proof with |oldSize| =
+		// 0 is empty. With |oldSize| = 1 (or |oldSize| = 2) it
+		// doesn't include a hash that depends on entry 0 (resp.
+		// 0 or 1). We ensured newSize > 3 when we assigned this
+		// worker its job, so we can use oldSize = 3.
+		oldSize = 3
```
mozkeeler, line 573: Ok - so `oldSize` will never be `0` in `consistencyProofToSubtrees`.

go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -79,6 +82,167 @@ type CtLogEntry struct {
...
+
+func (v *CtLogSubtreeVerifier) CheckClaim() error {
+	if len(v.Subtree.Root) != crypto.SHA256.Size() {
+		return fmt.Errorf("CtLogSubtreeVerifier: Claim has the wrong length.")
+	}
+	if v.numConsumed != v.Subtree.Size() {
+		return fmt.Errorf("CtLogSubtreeVerifier: Consumed %d leaves but needed %d.", v.numConsumed, v.Subtree.Size())
+	}
+	// Finish computing the Merkle tree head.
+	for len(v.hashStack) > 1 {
```
mozkeeler, line 159: Yeah, I feel like `Consume` should be responsible for this.

go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -278,100 +448,240 @@ func (ld *LogSyncEngine) NewLogWorker(ctLogUrl string) (*LogWorker, error) {
...
+		// until there are at least 3 entries in the log. So
+		// we'll wait.
+		task = Sleep
+	} else if logObj.LastUpdateTime.IsZero() {
+		// First contact with log
+		task = Init
+	} else if logObj.MaxEntry < sth.TreeSize-batchSize {
+		// There are many new entries to download.
+		task = Update
+	} else if logObj.MinEntry > 0 {
```
mozkeeler, line 465: Is `MinEntry > 0` right here? Won't that be true almost always, after we've done some downloading?

mozkeeler (approved):
Looks good - I left a couple of comments.

go/cmd/ct-fetch/ct-fetch.go
jschanck: Right, thanks for checking.

go/cmd/ct-fetch/ct-fetch.go
jschanck: It downloads in reverse order, so MinEntry > 0 is true until we've read all the way back to the beginning of the log.

mozkeeler: Ohh, right - ok.

go/cmd/ct-fetch/ct-fetch.go
jschanck: I agree. I tacked the change onto Part 4: 29c1c3cc13fd7dae7ad35cacfb8e1ef9a81feb16

mozkeeler: 👍

PR #192, opened 2021-12-09
Part 3 of ct-fetch upgrade

jschanck:
The first 8 commits are from parts 1 and 2. The rest allow us to configure the set of enrolled logs in remote settings.

* 255b189d070ca14c79c021539be002f3fa70c072 Replaces the URL parameter that we previously identified logs by with a collection of other metadata.
* 90550fb45b6610896924a905c3335f4470adf0c1 Saves more metadata about enrolled logs in redis. This metadata is read out by the aggregate jobs, which will be upgraded in a separate PR. 
* 34d2f764f293139e60c23f54458ec31a63a7c57c Loads the enrolled log list from remote settings, but only once at startup.
* adf288c3b2478225e88cc8da182bc6912ebff935 Adds support for periodically fetching the enrolled log list from remote settings.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -34,6 +34,7 @@ type CTConfig struct {
 	StatsDHost         *string
 	StatsDPort         *int
 	HealthAddr         *string
+	RemoteSettingsUpdateInterval *uint64
 }
 
 func confInt(p *int, section *ini.Section, key string, def int) {
@@ -139,6 +140,7 @@ func NewCTConfig() *CTConfig {
 		SavePeriod:         new(string),
 		StatsRefreshPeriod: new(string),
 		PollingDelay:       new(uint64),
+		RemoteSettingsUpdateInterval:       new(uint64),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


go/config/config.go
```diff
@@ -180,6 +182,7 @@ func (c *CTConfig) Init() {
 	confBool(c.LogExpiredEntries, section, "logExpiredEntries", false)
 	confBool(c.RunForever, section, "runForever", false)
 	confUint64(c.PollingDelay, section, "pollingDelay", 600)
+	confUint64(c.RemoteSettingsUpdateInterval, section, "remoteSettingsUpdateInterval", 60)
```
mozkeeler, line 185: 60 seconds seems a bit frequent for this.

mozkeeler (approved):

go/config/config.go
jschanck: haha, yeah there's a commit in part 4 that changes it to an hour. I had it at 60s for manual testing.

PR #193, opened 2021-12-09
Part 4 of ct-fetch upgrade

jschanck:
Another grab bag of changes. The most important is 3470cba507c570cb73d3865c8e12b8562077d6ea, which uses go context cancellation to try and cleanly shut down worker threads. Mostly a stylistic change, we were previously listening for signals in each worker thread and this just puts the signal handling logic in one place.

### DIFF
--- a/go/config/config.go
+++ b/go/config/config.go
@@ -21,8 +21,7 @@ type CTConfig struct {
 	GoogleProjectId     *string
 	RedisHost           *string
 	RedisTimeout        *string
-	Offset              *uint64
-	Limit               *uint64
+	BatchSize           *uint64
 	NumThreads          *int
 	RunForever          *bool
 	IssuerCNFilter      *string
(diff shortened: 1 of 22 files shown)


go/config/config.go
```diff
@@ -180,7 +180,7 @@ func (c *CTConfig) Init() {
 	confBool(c.LogExpiredEntries, section, "logExpiredEntries", false)
 	confBool(c.RunForever, section, "runForever", false)
 	confUint64(c.PollingDelay, section, "pollingDelay", 600)
-	confUint64(c.RemoteSettingsUpdateInterval, section, "remoteSettingsUpdateInterval", 60)
+	confUint64(c.RemoteSettingsUpdateInterval, section, "remoteSettingsUpdateInterval", 3600)
```
mozkeeler, line 183: Oh, haha - guess I should have looked at this first. An hour seems more reasonable.

mozkeeler (approved):
Looks good! I think this whole stack of PRs might need a `go fmt`?

PR #195, opened 2021-12-10
aggregate-known: save ct log metadata

### LINKED ISSUE #194: aggregate-known should save ct-log metadata
Aggregate-known should store the ct-log metadata that ct-fetch has collected. And it should do this prior to pulling the list of known certs from redis.

jschanck:
The `ct-fetch` job now saves metadata in redis about which log entries it has seen. With this PR, `aggregate-known` will read the CT log metadata from redis and save it in a "ct-logs.json" file. It's important that `aggregate-known` performs this task, and not `moz_kinto_publisher`, because ct-fetch runs continuously and we want the metadata at the time when `aggregate-known` was called.
 
Resolves #194


mozkeeler (approved):
LGTM.

PR #197, opened 2021-12-10
moz_kinto_publisher: Publish filter coverage

### LINKED ISSUE #196: moz_kinto_publisher should publish `coverage` information
I've updated the schema for full filters in the `cert-revocations` collection with a `coverage` field.

The coverage information is a list of CT logs and corresponding windows in time. If a certificate from an enrolled issuer has an embedded SCT with a timestamp from any of the given logs, and that timestamp is in the given window, then the filter is guaranteed to cover the certificate.

The `moz_kinto_publisher` job should be updated to populate this field.

jschanck:
Resolves #196

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -542,15 +542,18 @@ def publish_intermediates(*, args, rw_client):
             log.error(f"Couldn't unenroll record id {record}: {ke}")
 
     # Delete any remote records that had parsing errors
-    for record in remote_error_records:
-        log.info(f"Deleting remote record with parsing error: {record}")
+    # (note these "records" are just dictionaries)
+    for raw_record in remote_error_records:
+        log.info(f"Deleting remote record with parsing error: {raw_record}")
         try:
             rw_client.delete_record(
                 collection=settings.KINTO_INTERMEDIATES_COLLECTION,
-                id=record["id"],
+                id=raw_record["id"],
             )
         except KintoException as ke:
-            log.error(f"Couldn't delete record id {record['id']}: {ke}")
+            log.error(f"Couldn't delete record id {raw_record['id']}: {ke}")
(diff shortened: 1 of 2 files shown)


moz_kinto_publisher/main.py
```diff
@@ -553,10 +559,10 @@ def publish_intermediates(*, args, rw_client):
         try:
```
mozkeeler, line 559: There's a line about 10 up that also uses `record["id"]` to delete an intermediate (thanks for not letting me comment on it, github 🙄) - does that need to change too?

moz_kinto_publisher/main.py
```diff
@@ -646,85 +652,101 @@ def clear_crlite_stashes(*, rw_client, noop):
...
             collection=settings.KINTO_CRLITE_COLLECTION,
-            data=attributes,
-            permissions=perms,
+            fileName=attachment_name,
+            filePath=attachment_path,
+            recordId=recordid,
+        )
+    except KintoException as ke:
+        log.error(
+            f"Failed to upload attachment. Removing stale MLBF record {recordid}: {ke}"
```
mozkeeler, line 682: This is also used for stashes, right? Let's replace `MLBF` with `CRLite` or something more generic.

mozkeeler (approved):
Seems reasonable. I left a couple of comments.

moz_kinto_publisher/main.py
jschanck: The one above is right, and it's the reason I initially made the type error here. The one above is an error case where we didn't create an `Intermediate` object, so we only have a dictionary. I cleaned it up a bit in the most recent commit.

moz_kinto_publisher/main.py
jschanck: Done

Issue #207, opened 2021-12-15
meta: review log messages


jschanck:
CRLite produces a lot of log messages. We should review the messages and their severity levels to make it easier to identify important messages at a glance.

PR #208, opened 2021-12-15
moz_kinto_publisher: avoid a confusing log message in dev environment

jschanck:
Calling `request_review_of_collection` in the dev environment results in an HTTP error because we don't have access to the right server. We can avoid this by skipping the call if `noop` is enabled

Part of the general log message cleanup in #207


mozkeeler (approved):
Seems reasonable.

PR #210, opened 2021-12-15
crlite-signoff-tool: restore previous behavior for KINTO_NOOP

### LINKED ISSUE #205: Restore behavior of KINTO_NOOP in crlite-signoff-tool
The crlite-signoff.sh script was using the convention that the `--noop` flag for crlite-signoff-tool.py would be set if the `KINTO_NOOP` environment variable was set. It did not inspect the value of `KINTO_NOOP` if it was set. In cc90c3bc711e866c8b8bb3d2614677acbad000e5 I changed crlite-signoff-tool.py so it would check the `KINTO_NOOP` flag itself, but inadvertently changed the behavior so that `--noop` would be set only if `KINTO_NOOP=True`. The original behavior should be restored.

jschanck:
Resolves #205


mozkeeler (approved):
Ok

PR #212, opened 2021-12-15
moz_kinto_publisher: skip ctlog data with empty LogID

### LINKED ISSUE #209: moz_kinto_publisher should ignore data from old versions of ct-fetch when producing coverage metadata
Old versions of ct-fetch produce log metadata without a "LogID" field. The moz_kinto_publisher script should ignore these entries rather than publishing (useless) coverage metadata.

jschanck:
Resolves #209


mozkeeler (approved):
Makes sense.

PR #215, opened 2021-12-16
moz_kinto_publisher: exit if most recent run is not ready

### LINKED ISSUE #214: moz_kinto_publisher should exit if the most recent run is not ready
There's no reason for moz_kinto_publisher to run through consistency checks if we time out waiting for a run to complete.

jschanck:
If we don't exit when the most recent run is not ready we'll fail in a later consistency check. This produces some confusing log messages.

Resolves #214


mozkeeler (approved):
Sounds good.

PR #217, opened 2021-12-17
moz_kinto_publisher: adjust coverage by MMD

### LINKED ISSUE #216: moz_kinto_publisher should be adjust coverage by MMD
The moz_kinto_publisher script currently puts the exact values of `MinTimestamp` and `MaxTimestamp` that it receives from ct-fetch into each log's `coverage` metadata. I had originally planned to have clients adjust these values by the log's MMD before doing a coverage check, as that's similar to the existing behavior (clients are currently adjusting the filter timestamp by MMD in `NSSCertDBTrustDomain::CheckCRLite`).

On reflection I think moz_kinto_publisher should adjust the timestamps before including them in the `coverage` metadata. There are a couple of special cases to handle when ct-fetch has a "complete" log, and I'd rather not push that logic onto clients.

jschanck:
Resolves #216

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -733,6 +733,32 @@ def publish_crlite_main_filter(
     record_epoch_time_ms = math.floor(timestamp.timestamp() * 1000)
     identifier = f"{record_time}Z-full"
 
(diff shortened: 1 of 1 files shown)


mozkeeler (approved):
LGTM.

PR #223, opened 2022-01-12
Final changes for v1.0.6

### LINKED ISSUE #218: Update minimum required version of moz_crlite_query
The moz_crlite_query script has been updated to respect coverage metadata. Once a new version is released we need to bump the minimum required version here.

### LINKED ISSUE #220: Wrong units/scaling for time.Duration lead to liveness probe failures
Our staging instance of ct-fetch is failing liveness probe tests because the maximum time since last update is set too small.

### LINKED ISSUE #222: Avoid duplicate log::url keys in redis
In older versions of ct-fetch it's possible to create a pair of entries in redis with keys `log::example.com` and `log::example.com/`. Newer versions omit the slash. Our production instance of ct-fetch has some keys with the slash. We should migrate these entries on upgrade.

jschanck:
Commit ad65b2a resolves #218. The three commits that follow are minor changes. Commit a2c5f29 resolves #220. Commit 7432053 resolves #222.

### DIFF
--- a/go/cmd/ct-fetch/ct-fetch.go
+++ b/go/cmd/ct-fetch/ct-fetch.go
@@ -329,6 +329,9 @@ func (ld *LogSyncEngine) SyncLog(ctx context.Context, enrolledLogs *EnrolledLogs
 			return err
 		}
 
+		// We did useful work. Register an update for the health service.
+		ld.RegisterUpdate()
+
 		if !*ctconfig.RunForever {
 			return nil
 		}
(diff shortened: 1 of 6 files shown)


go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -520,12 +531,13 @@ func (ld *LogSyncEngine) NewLogWorker(ctx context.Context, ctLogData *types.CTLo
 
 func (lw *LogWorker) sleep(ctx context.Context) {
 	// Sleep for ctconfig.PollingDelay seconds (+/- 10%).
-	duration := time.Duration((1 + 0.1*rand.NormFloat64()) * float64(*ctconfig.PollingDelay))
```
mozkeeler, line 523: oof

mozkeeler (approved):
Looks good to me.

go/cmd/ct-fetch/ct-fetch.go
jschanck: That one was just cleanup, I was multiplying by time.Second in line 535. The bad one was `evaluationTime` in the health check. That one was causing the ct-fetch process to be rebooted every 7 minutes or so...

PR #226, opened 2022-01-25
aggregate-crls: enroll issuers with accessible but empty CRLs

### LINKED ISSUE #225: Enroll issuers with functioning but empty CRLs
We currently only enroll an issuer if a) we've successfully downloaded all of their CRLs, and b) they've revoked at least one certificate through those CRLs. We should remove the second condition, as it causes unnecessary churn in the enrollment list.

jschanck:
Resolves #225


mozkeeler (approved):
Makes sense.

PR #228, opened 2022-01-25
moz_kinto_publisher: Update enrollment only when a full filter is published 

### LINKED ISSUE #227: Update enrollment only when a full filter is published
To reduce the risk that clients' issuer enrollment list gets out of sync with the published filter, we should only update enrollment when a full filter is published. We should also check that the enrollment status in the `intermediates` collection matches what's in cloud storage as part of the sign-off process. Additional changes will be made client-side under [Bug 1750787](https://bugzilla.mozilla.org/show_bug.cgi?id=1750787).

jschanck:
Resolves #227

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -499,14 +499,21 @@ def load_remote_intermediates(*, kinto_client):
     return remote_intermediates, remote_error_records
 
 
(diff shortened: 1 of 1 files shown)


mozkeeler (approved):
Seems reasonable.

PR #237, opened 2022-02-15
Delete remote intermediates not on local list

### LINKED ISSUE #236: Intermediates collection should be a subset of the ICAs disclosed in CCADB
Every entry in the `intermediates` collection in Remote Settings was, at one time, on a list provided to us by CCADB. Sometimes an intermediate drops off this list, but we keep the corresponding record in Remote Settings. This might be useful for intermediates preloading, but it's currently complicating the signoff logic for CRLite. We should remove entries from `intermediates` which are no longer on the list from CCADB.

jschanck:
Resolves #236 

Note: `remote_expired` constructed on line 555 is a subset of `remote_only`. So there's no need to delete expired certs in a separate loop.

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -571,19 +571,19 @@ def publish_intermediates(*, args, rw_client, new_filter_run_id=None):
         log.info("Noop flag set, exiting before any intermediate updates")
         return
 
-    # Enrolled intermediates must be in the local list
-    for unique_id in remote_only & remote_enrolled:
+    # All intermediates must be in the local list
+    for unique_id in remote_only:
         record = remote_intermediates[unique_id]
-        log.info(f"Unenrolling deleted {record} from CRLite")
+        log.info(f"Removing deleted intermediate {record}")
         try:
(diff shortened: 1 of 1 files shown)


moz_kinto_publisher/main.py
```diff
@@ -594,18 +594,6 @@ def publish_intermediates(*, args, rw_client, new_filter_run_id=None):
         except KeyError:  # raw_record doesn't have "id"
             log.error(f"Couldn't delete record: {raw_record}")
 
-    # Delete any expired remote records
-    for unique_id in remote_expired:
```
mozkeeler, line 598: So is `remote_expired` used any longer?

mozkeeler (changes requested):
We don't need `remote_expired` any longer, right?

moz_kinto_publisher/main.py
jschanck: Yes, we log the number of expired certs on line 565. I thought about deleting it, but it's nice to have an explanation for some removals.

mozkeeler: Okay, sounds good.

mozkeeler (approved):

PR #240, opened 2022-02-19
Relax sign-off checks for unenrolled intermediates

### LINKED ISSUE #239: Relax sign-off checks for unenrolled intermediates
In v1.0.8 we added sign-off checks to ensure that the enrollment list at remote settings matched the enrollment list for the current full filter. There are a few situations where the new checks are too strict.
1) We might want to add a new intermediate in the `crlite_enrolled = False` state between full filter updates. These are useful for preloading.
2) We might want to unenroll an issuer from CRLite (because of misbehavior of some kind) without pushing a full filter update.

We should relax the signoff checks to allow remote settings to differ from the aggregator's enrollment list for intermediates in the `crlite_enrolled = False` state.

jschanck:
Resolves #239 

Also fixes a confusing log message in moz_kinto_publisher.

### DIFF
--- a/containers/scripts/crlite-signoff-tool.py
+++ b/containers/scripts/crlite-signoff-tool.py
@@ -127,7 +127,7 @@ class SignoffClient(Client):
                 log.warning(
                     f"Missing remote settings record for {r['subject']}--{r['pubKeyHash']}"
                 )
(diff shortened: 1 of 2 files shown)


mozkeeler (approved):
Ok - makes sense.

PR #242, opened 2022-02-25
Store the current filter's enrolled issuer list in cert-revocations

### LINKED ISSUE #241: Store the current filter's enrolled issuer list in cert-revocations
We're going to move the enrollment list from the `intermediates` collection to the `cert-revocations` collection.

The schema for `cert-revocations` has been updated with an `enrolledIssuers` field that holds a JSON array of strings. Each string in `enrolledIssuers` is to be a base64 encoded identifier for a CRLite-enrolled issuer. The identifier will be `SHA256(subject || spki)` where `subject` is a DER encoded RFC 5280 `RDNSequence` and `spki` is a DER encoded RFC 5280 `SubjectPublicKeyInfo`.

The CRLite aggregator should output the new identifiers in `enrolled.json`, and `moz_kinto_publisher` should populate the `enrolledIssuers` field. The signoff script should validate the `enrolledIssuers` field.

We'll continue to keep the `crlite_enrolled` flags in `intermediates` up to date for now.

Client-side changes will be tracked in [Bug 1750787](https://bugzilla.mozilla.org/show_bug.cgi?id=1750787)

jschanck:
Resolves #241 

~Note: The `crlite_enrollment_id.py` script is not used here. I needed it while making tests for the client-side code, and I think it makes more sense to keep it here with CRLite than mozilla-central.~ [edit: moved script to mozilla-central]

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -870,6 +870,8 @@ def crlite_verify_record_consistency(*, existing_records):
             raise ConsistencyException(f"Malformed record {r}.")
         if not r["incremental"] and not "coverage" in r:
             raise ConsistencyException(f"Malformed record {r}.")
+        if not r["incremental"] and not "enrolledIssuers" in r:
+            raise ConsistencyException(f"Malformed record {r}.")
 
     # There must be exactly 1 full filter in the existing records.
     full_filters = [r for r in existing_records if not r["incremental"]]
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -738,7 +738,7 @@ def publish_crlite_record(
 
 
 def publish_crlite_main_filter(
-    *, rw_client, filter_path, filename, timestamp, ctlogs, noop
+    *, rw_client, filter_path, filename, timestamp, ctlogs, intermediates, noop
 ):
     record_time = timestamp.isoformat(timespec="seconds")
     record_epoch_time_ms = math.floor(timestamp.timestamp() * 1000)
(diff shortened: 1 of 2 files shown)


util/crlite_enrollment_id.py
```diff
@@ -0,0 +1,33 @@
...
+if len(sys.argv) != 2:
+    print(f"Usage: {sys.argv[0]} <path to pem cert>")
+    sys.exit(1)
+
+with open(sys.argv[1], "r") as f:
+    cert = x509.load_pem_x509_certificate(f.read().encode("utf-8"))
+
+subj = cert.subject.public_bytes()
+spki = cert.public_key().public_bytes(
+    format=serialization.PublicFormat.SubjectPublicKeyInfo,
```
mozkeeler, line 26: I'm a little concerned that we can't directly use the bytes of the certificate as-is, but I think if there's an encoding mismatch (which there shouldn't be, but, well...), we'll fall back to OCSP fetching, so this should be fine.

util/crlite_enrollment_id.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/python
```
mozkeeler, line 1: Like I said in the other review, this should live where it's actually used. There's precedent for having utility scripts alongside tests in mozilla-central.

mozkeeler (approved):
Looks good. Just a couple of minor comments.

moz_kinto_publisher/main.py
```diff
@@ -799,11 +799,18 @@ def publish_crlite_main_filter(
             }
         ]
 
+    enrolledIssuers = []
+    for issuer in intermediates:
+        if issuer["enrolled"]:
+            uid = base64.urlsafe_b64decode(issuer["uniqueID"])
+            enrolledIssuers.append(base64.b64encode(uid))
```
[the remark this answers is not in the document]
jschanck: The aggregator-produced `enrolled.json` file uses urlsafe base64 for other values, so I did this for consistency. I'll open a separate bug to switch everything to standard base64; it's been on my TODO list.

util/crlite_enrollment_id.py
jschanck: This script is only used to generate test cases. The client and the aggregator get the bytes directly with moz::pkix and Go x509 respectively.

util/crlite_enrollment_id.py
jschanck: OK, I'll put a copy there.

mozkeeler (approved):
Sounds good!

PR #244, opened 2022-02-28
moz_kinto_publisher: add 'enrolledIssuers' field to test records

jschanck:
Fixes broken test on main.


mozkeeler (approved):
Yep - this looks like the right sort of thing.

PR #245, opened 2022-03-01
Fix issues with enrolledIssuers

jschanck:
Fix up an encoding issue, and only publish one copy of each issuer ID (`enrolled.json` has duplicates from cross-signed certs).


mozkeeler (approved):
Ah, yes, b64encode returns bytes. That's just good API design.

PR #252, opened 2022-04-14
Format with black

jschanck:


mozkeeler (approved):
Ah, yes, spaces around operators, but not these operators :D

PR #254, opened 2022-05-06
Use rust-cascade for filter generation

### LINKED ISSUE #253: Use rust-cascade for filter generation

jschanck:
Resolves #253


jschanck:
The way I'm sending filters to different channels here isn't actually going to work. I'll revise.

jschanck:
Skipping channelization for now, as we discussed. This will continue to generate murmurhash3 filters until we remove the `--murmurhash3` from `workflow/1-generate_mlbf`.

rust-create-cascade/src/main.rs
```diff
@@ -0,0 +1,525 @@
...
+
+    info!("Verifying cascade");
+    check_all(&cascade, revoked_dir, known_dir);
+
+    info!("Testing serialization");
+    let cascade_bytes = cascade.to_bytes().expect("cannot serialize cascade");
+    if let Some(cascade) =
+        Cascade::from_bytes(cascade_bytes.clone()).expect("cannot deserialize cascade")
+    {
+        info!("\n{}", cascade);
```
mozkeeler, line 322: Do we want to run `check_all` here again just to be sure? Or is that overkill?

rust-create-cascade/src/main.rs
```diff
@@ -0,0 +1,525 @@
...
+///
+fn write_revset_and_stash(
+    revset_file: &Path,
+    stash_file: &Path,
+    prev_revset_file: &Path,
+    revoked_dir: &Path,
+    known_dir: &Path,
+) {
+    let mut prev_keys: HashSet<Vec<u8>> = HashSet::new();
+    if prev_revset_file.exists() {
```
mozkeeler, line 370: Would we want to warn/info if this file doesn't exist?

rust-create-cascade/src/main.rs
```diff
@@ -0,0 +1,525 @@
...
+        err = true;
+    }
+    if !revoked_dir.exists() {
+        error!("{} not found", revoked_dir.display());
+        err = true;
+    }
+
+    if std::fs::create_dir_all(&out_dir).is_err() {
+        error!("Could not create out directory: {}", out_dir.display());
+        err = true
```
mozkeeler, line 474: nit: presumably meant to end this line with a `;`?

mozkeeler (approved):
Awesome! I just left a few minor comments.

rust-create-cascade/src/main.rs
```diff
@@ -325,6 +326,9 @@ fn create_cascade(out_file: &Path, revoked_dir: &Path, known_dir: &Path, hash_al
         return;
     }
 
+    info!("Verifying cascade");
+    check_all(&cascade, revoked_dir, known_dir);
```
mozkeeler, line 330: I meant additionally checking the cascade that gets read from `cascade_bytes`, in the `if` block, but it's probably not a big deal.

jschanck: Ack, that's what I meant to do. Good catch!

PR #261, opened 2022-05-25
Replace moz_crlite_query with rust-query-crlite

### LINKED ISSUE #260: Replace moz_crlite_query
`moz_crlite_query` isn't quite fitting our needs for several reasons:
1) it doesn't store filters/metadata in the same format as a firefox `security_state` directory, and
2) it doesn't support the new SHA256-based filters, and
3) it's written in python and cannot share code with rust-cascade or rust-create-cascade.

While we could upgrade `moz_crlite_query` to fix 1 and 2, but all together it makes sense to write a new query tool in rust.

jschanck:
Resolves #260 

Long term, `rust-query-crlite` should move to its own repo and we should break out a small library that can de-duplicate the parts of `rust-query-crlite` that are copied from firefox `cert_storage`.


rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+const PROD_ATTACH_URL: &str = "https://firefox-settings-attachments.cdn.mozilla.net/";
+const PROD_URL: &str = "https://firefox.settings.services.mozilla.com/v1/buckets/security-state/collections/cert-revocations/records";
+
+const COVERAGE_SERIALIZATION_VERSION: u8 = 1;
+const COVERAGE_V1_ENTRY_BYTES: usize = 48;
+
+const ENROLLMENT_SERIALIZATION_VERSION: u8 = 1;
+const ENROLLMENT_V1_ENTRY_BYTES: usize = 32;
+
+const OID_SCT_EXTENSION: &der_parser::Oid = &oid!(1.3.6 .1 .4 .1 .11129 .2 .4 .2);
```
mozkeeler, line 51: Does the macro require the spaces? I'm used to seeing OIDs without spaces.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+    let mut hasher = Sha256::new();
+    hasher.update(issuer_dn);
+    hasher.update(issuer_spki);
+    hasher.finalize().to_vec()
+}
+
+// x509_parser parses the serial number to a BigUint, but DER integers are signed
+// so we need to retain a leading 0 byte in some cases where BitUint::to_bytes_be
+// drops it.
+fn serial_bytes(n: &num_bigint::BigUint) -> Vec<u8> {
```
mozkeeler, line 76: We've had problems in the past with improperly encoded serial numbers (e.g. a leading 0 byte should be present but isn't). Is there a way to directly use the encoded bytes of the serial number (minus the leading integer tag and length byte(s))?

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+extern crate base64;
+extern crate bincode;
+extern crate byteorder;
+extern crate clap;
+extern crate der_parser;
+extern crate hex;
+extern crate log;
+extern crate num_bigint;
+extern crate pem;
+extern crate reqwest;
```
mozkeeler, line 12: uwu

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+        .map_err(|_| CRLiteDBError::from("could not fetch remote settings collection"))?
+        .json()
+        .map_err(|_| CRLiteDBError::from("could not read remote settings data"))?;
+
+    let (stashes, full_filters): (Vec<&JsonRecord>, Vec<&JsonRecord>) =
+        records.data.iter().partition(|x| x.incremental);
+
+    if full_filters.len() != 1 {
+        return Err(CRLiteDBError::from(
+            "multiple full filters in remote settings",
```
mozkeeler, line 176: Well, or 0.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+    }
+
+    // Skip stash update if the filter is fresh and the stash on disk has the expected size.
+    // (We can't easily check the provided hashes since we concatenate stashes on disk.)
+    let expected_stash_size = stashes.iter().fold(0, |x, y| x + y.attachment.size);
+    let stash_metadata = std::fs::metadata(&stash_path);
+    if !filter_needs_update
+        && stash_path.exists()
+        && stash_metadata.is_ok()
+        && stash_metadata.unwrap().len() == expected_stash_size
```
mozkeeler, line 203: I think you could use `map_or` or something to avoid the `is_ok` and `unwrap`.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+        }
+        Ok(Stash(stash))
+    }
+}
+
+struct Coverage(HashMap<LogID, (u64, u64)>);
+impl Coverage {
+    fn from_bytes(bytes: &[u8]) -> Result<Self, CRLiteDBError> {
+        let mut reader = BufReader::new(bytes);
+        if (bytes.len() - 1 % COVERAGE_V1_ENTRY_BYTES) == 0 {
```
mozkeeler, line 416: Should this be `!= 0`?

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+        }
+        let enrollment_count = (bytes.len() - 1) / ENROLLMENT_V1_ENTRY_BYTES;
+        for _ in 0..enrollment_count {
+            let mut enrollment_entry = [0u8; ENROLLMENT_V1_ENTRY_BYTES];
+            match reader.read_exact(&mut enrollment_entry) {
+                Ok(()) => (),
+                _ => return Err(CRLiteDBError::from("truncted CRLite enrollment file")),
+            };
+            let issuer_id = &enrollment_entry[..];
+            enrollment.insert(issuer_id.to_vec());
```
mozkeeler, line 490: We can avoid `issuer_id` altogether and just say `enrollment_entry.to_vec()` I think.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+        //      the logical line is split over several actual lines. Fortunately the pem crate is
+        //      happy to ignore content surrounding PEM data.
+        let list = pem::parse_many(bytes)
+            .map_err(|_| CRLiteDBError::from("error reading CCADB report"))?;
+
+        let mut intermediates = Intermediates::new();
+        for der in list {
+            if let Ok((_, cert)) = X509Certificate::from_der(&der.contents) {
+                let name = cert.tbs_certificate.subject.as_raw();
+                if let Some(existing) = intermediates.0.get_mut(name) {
```
mozkeeler, line 518: Could you do `intermediates.0.entry(name).or_insert_with(Vec::new)` and then append on to the result of that? Not a big deal, but it's similar to what you do above.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+                .map_err(|_| CRLiteDBError::from(format!("invalid DNS name: {}", host)))?;
+
+            let mut conn = rustls::ClientConnection::new(Arc::clone(&config), server_name)
+                .map_err(|e| CRLiteDBError::from(format!("{}: tls error: {}", host, e)))?;
+
+            let mut sock = TcpStream::connect(addr)
+                .map_err(|e| CRLiteDBError::from(format!("{}: tcp error: {}", host, e)))?;
+
+            let mut tls = rustls::Stream::new(&mut conn, &mut sock);
+            let _ = tls.write_all("HEAD / HTTP/1.1\r\n\r\n".as_bytes());
```
mozkeeler, line 617: Maybe include a `X-Automated-Tool` header?

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+                        Status::Expired => warn!("{} {:?}", host, status),
+                        Status::Good => info!("{} {:?}", host, status),
+                        Status::NotCovered => warn!("{} {:?}", host, status),
+                        Status::NotEnrolled => warn!("{} {:?}", host, status),
+                        Status::Revoked => {
+                            found_revoked_certs = true;
+                            error!("{} {:?}", host, status);
+                        }
+                    }
+                    continue 'host_loop; // no need to check further addrs
```
mozkeeler, line 633: Rather than have a big outer loop, maybe there could be a function to check just one host at a time.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+
+fn query_certs(db: &CRLiteDB, files: &[PathBuf]) -> Result<CmdResult, CRLiteDBError> {
+    let mut found_revoked_certs = false;
+    for file in files {
+        if !file.exists() {
+            warn!("File does not exist: {}", file.display());
+            continue;
+        }
+        let input = std::fs::read(&file);
+        if input.is_err() {
```
mozkeeler, line 657: I think you can
```
let input = match input {
  Ok(input) => input,
  Err(_) => continue,
};
```
to avoid the `is_err` and `unwrap`.

rust-query-crlite/src/main.rs
```diff
@@ -0,0 +1,794 @@
...
+        error!("CRLite DB is incomplete. Use --update [prod | stage] to populate");
+        std::process::exit(1);
+    }
+
+    let db = CRLiteDB::load(&args.db);
+    if let Err(e) = db {
+        error!("Error loading CRLite DB: {}", e.message);
+        std::process::exit(1);
+    }
+    let db = db.unwrap();
```
mozkeeler, line 775: I thiiiink you can do a match again to avoid the unwrap, but I don't know if the compiler is smart enough to know that the exit doesn't continue code flow.

mozkeeler (changes requested):
Awesome! I made a few minor notes.

rust-query-crlite/src/main.rs
jschanck: I left this one as is. If we `map_or` then we need a default value that is guaranteed not to equal `expected_stash_size`. We could use `u64::MAX`, but I prefer to test `is_ok`.

rust-query-crlite/src/main.rs
jschanck: Yes! and the bracketing is wrong.

mozkeeler (approved):
LGTM!

PR #264, opened 2022-06-15
rust-query-crlite: handle negative serial numbers

jschanck:
I overlooked the `raw_serial()` method on `x509_parser::TbsCertificate`. Under the hood it calls `der_parser::parse_ber_integer` https://docs.rs/der-parser/latest/src/der_parser/ber/parser.rs.html#280-295, which is exactly what we want.


mozkeeler (approved):
Great!

PR #267, opened 2022-10-04
ct-fetch: include CA certs

### LINKED ISSUE #266: Include CA certs in filters
Firefox doesn't query CRLite on CA certs, but other applications might. We don't communicate the fact that we're not including CA certs in our filter metadata, which means other applications could get false positives if they query CA certs.

In the interest of making our filters broadly usable, let's include CA certs. The overall impact on filter size will be small.

jschanck:
Resolves #266


mozkeeler (approved):
Seems reasonable

PR #271, opened 2022-10-12
aggregate-crls: Use CRLs from CCADB

### LINKED ISSUE #270: Use CRLs from CCADB report
In addition to discovering CRLs in certificates from CT, we should pull CRLs from the CCADB report.

jschanck:
Resolves #270 

Note that this does not yet include a countermeasure against the attack [described recently on dev-security-policy](https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/qhrGxLvyreU). We can revisit this if/when the policy is updated with a countermeasure.


mozkeeler (approved):
Looks good to me!

PR #274, opened 2022-10-20
Reduce the number of updates we make to `intermediates`

### LINKED ISSUE #272: Normalize PEM data from CCADB report
The CCADB report often includes PEM data with misplaced line breaks, etc. There are several places in the code that use a hash of PEM data for caching, and we sometimes suffer cache misses because of the formatting errors in the CCADB report. We should normalize PEMs in `root_program/issuers.go` before writing the issuer list.

### LINKED ISSUE #273: Stop tracking CRLite enrollment in intermediates collection
CRLite clients no longer use the `crlite_enrolled` flag in the intermediates collection. We can stop updating it.

jschanck:
The first patch normalizes the PEM encoding of certificates in `enrolled.json`. If we made this change alone there would be some 600 unnecessary attachment replacements in `intermediates`. The third patch changes how we decide when to upload a new attachment so that we will not make these replacements.

The second patch will cause us to set all of the `crlite_enrolled` flags to false in `intermediates`. This will have a one-time cost of causing clients to re-download the affected ~800 records (but not their attachments). There are many intermediates that get enrolled and then unenrolled from CRLite every few weeks due to the availability of their CRLs. So long term this change should reduce the load on `intermediates`.

Resolves #272
Resolves #273


mozkeeler (approved):
Looks good to me.

PR #275, opened 2022-10-20
ct-fetch: keep certificates with non-fatal parsing errors

### LINKED ISSUE #174: CRLite shouldn't drop certs for non fatal parsing errors
CRLite shouldn't drops certificates with non fatal parsing errors.

Here's an example log message:
> "[https://ct.cloudflare.com/logs/nimbus2023/] Problem decoding certificate: index: 11300 error: NonFatalErrors: asn1: syntax error: Value: PrintableString contains invalid character; "

Here's the corresponding cert: https://crt.sh/?id=4019855673&opt=x509lint
Here's where we drop it: https://github.com/mozilla/crlite/blob/3bdafdc6796b020e8983b5f1e4694a32b1d13c72/go/cmd/ct-fetch/ct-fetch.go#L207

jschanck:
Resolves #174


go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -393,7 +393,7 @@ func (ld *LogSyncEngine) insertCTWorker() {
 			precert = true
 		}
 
-		if err != nil {
```
mozkeeler, line 396: Should we still log `err` as a warning if it's non-nil, maybe?

mozkeeler (approved):
Sounds good.

go/cmd/ct-fetch/ct-fetch.go
jschanck: sure!

PR #277, opened 2022-10-27
ct-fetch: retry downloads for some network errors

### LINKED ISSUE #276: ct-fetch stalls after temporary network outage
Some errors, such as a timeout while requesting a signed tree head from a CT log, cause ct-fetch to assume that a log is inoperable and then to stop downloading. Let's instead have ct-fetch sleep and then retry.

jschanck:
Resolves #276


mozkeeler (approved):
Sounds good!

PR #279, opened 2022-10-28
ct-fetch: fix firstIndex for init job on CT log with few entries

### LINKED ISSUE #278: ct-fetch cannot download from logs with few entries
ct-fetch tries to download `JobSize` entries between a `firstIndex` and a `lastIndex`. If `JobSize` is greater than the total number of entries in the log, it suffers a uint64 underflow in the calculation of `firstIndex`. The log rightfully responds that there are no entries in the requested range. Mostly harmless, but we end up making an invalid request to a CT log roughly every 10 minutes.

jschanck:
Resolves #278


mozkeeler (approved):
Sounds good!

PR #281, opened 2022-11-02
aggregate-known: avoid keeping an issuer's full serial list in memory

### LINKED ISSUE #280: Reduce memory requirements of the aggregate-known task
The redis backend stores serials sharded by issuer and expiry date, but aggregate-known concatenates all serials from a single issuer into an in-memory buffer before writing them out to disk. Since enrolling Let's Encrypt, this has caused some OOM failures in our production instance. Let's stream the serials to disk as we read shards.

jschanck:
Resolves #280 

I tested this on dev. Here's the memory usage before (< 1pm) and after (> 1pm):
before_and_after

In the before graph, the drop around 12:30 is the end of aggregate-known, and the final spike before that is loading serials from Let's Encrypt. We have enough certs from LE now that we were getting consistent OOM failures there. 

In the after graph, the initial bump at 1:05 is in aggregate-crls. It has the same issue as aggregate-known, but the number of serials is much less. I'm going to avoid tinkering with it for now. The main point is that the memory usage in aggregate-known is negligible.

The memory usage in the last half hour of both images is when we're actually generating the filter. We haven't seen OOM issues there, but I'm running a heap profiler on a local instance to see if I can reduce the memory usage at all.

### DIFF
--- a/go/cmd/aggregate-known/aggregate-known.go
+++ b/go/cmd/aggregate-known/aggregate-known.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"flag"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


mozkeeler (approved):
Looks good to  me.

PR #282, opened 2023-01-04
signoff: Configure filter size limit through max_filter_size env var

jschanck:
This adds a bit more debugging info during signoff and makes the filter size cutoff configurable through an environment variable.


mozkeeler (approved):
SGTM!

PR #292, opened 2023-04-19
Use rust:latest in Dockerfile

jschanck:
Builds are currently failing with
> error: package `time v0.3.20` cannot be built because it requires rustc 1.63.0 or newer, while the currently active rustc version is 1.61.0

I don't see any harm in building with the `rust:latest` image.


mozkeeler (approved):
Ok!

PR #295, opened 2024-02-27
Ignore CRLs that are not in CCADB and update some dependencies

jschanck:
Now that CAs in our root program are required to publish CRLs in CCADB, we don't need to fetch every CRL that we find in CT.


mozkeeler (approved):
LGTM!

PR #296, opened 2024-02-28
Upload source data before generating mlbf for easier debugging

jschanck:
This splits `workflow/2-upload_artifacts_to_storage` into `workflow/1-upload_data_to_storage` and  `workflow/3-upload_mlbf-to_storage` and makes the former run before the filter generation workflow. The idea here is that if filter generation fails we want to be able to pull the source material from GCP for local debugging.

Also a bit of maintenance here to get a base docker image that can run binaries from recent versions of rust. And some changes to make the logs a bit quieter.


containers/scripts/crlite-generate.sh
```diff
@@ -43,13 +43,20 @@ ls -latS ${ID}/revoked | head
 echo "crlite-fullrun: disk usage"
 du -hc ${ID}
 
-${workflow}/1-generate_mlbf ${ID} \
+if [ "x${DoNotUpload}x" == "xx" ] ; then
+  echo "archiving crls"
+  tar -czf ${ID}/crls.tar.gz -C ${CRLS} .
+  echo "uploading source materials"
+  ${workflow}/1-upload_data_to_storage ${ID} \
```
mozkeeler, line 50: Should this be `1-upload_artifacts_to_storage`? Or is there a missing file or something?

hooks/go-no-replacement.sh
mozkeeler, line 1: I think this is still referenced in `[.pre-commit-config.yaml](https://github.com/mozilla/crlite/blob/f1ccc84b92866386ed15bbecf548d22f60c02733/.pre-commit-config.yaml#L30)` - should we just remove that too?

workflow/1-upload_artifacts_to_storage
mozkeeler, line 1: This seems to be largely the same as `[3-upload_mlbf_to_storage](https://github.com/mozilla/crlite/pull/296/files#diff-2af64200fdbc3cf8f17b66f51f3515de134fc381d689ecb1185e100ff63ad816)` - was this intended? Is there a way they could share code?

mozkeeler (changes requested):
I think there might be a missing file or something?

containers/scripts/crlite-generate.sh
jschanck: ah, yeah, hg <-> git context switch  mishap.

mozkeeler (approved):
LGTM! Just the one comment.

PR #298, opened 2024-02-29
Fetch previous revset.bin from latest folder

jschanck:
The change in v1.0.20 to upload source data before uploading the new filter breaks the mechanism that `workflow/2-generate-mlbf` used to find the previous run's `revset.bin` file. There's no reason to look up the latest run's identifier, we can just grab `revset.bin` from the `/latest` folder.

### DIFF
--- a/workflow/2-generate_mlbf
+++ b/workflow/2-generate_mlbf
@@ -43,20 +43,16 @@ def main():
 
     if not args.nodiff and "GOOGLE_APPLICATION_CREDENTIALS" in os.environ:
         try:
-            all_runs = workflow.get_run_identifiers(args.filter_bucket)
-            if all_runs:
-                latest = all_runs.pop()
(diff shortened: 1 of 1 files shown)


workflow/2-generate_mlbf
```diff
@@ -43,20 +43,16 @@ def main():
...
+            prev_revset = dest / Path("prev_revset.bin")
 
-                workflow.download_from_google_cloud(
-                    args.filter_bucket,
-                    f"{latest}/mlbf/revset.bin",
-                    prev_revset,
-                )
+            workflow.download_from_google_cloud(
+                args.filter_bucket,
+                f"latest/revset.bin",
```
mozkeeler, line 51: I guess technically we don't need the `f` prefix?

mozkeeler (approved):
Makes sense!

PR #300, opened 2024-03-05
Update rust-query-crlite dependencies

jschanck:
Update a couple of dependencies to pick up new signature algorithm support, etc.

### DIFF
--- a/rust-query-crlite/Cargo.toml
+++ b/rust-query-crlite/Cargo.toml
@@ -8,7 +8,7 @@ base64 = "0.13"
 bincode = "1.3"
 byteorder = "1.2.7"
 clap = { version = "3.0", features = ["derive"] }
-der-parser = "7.0"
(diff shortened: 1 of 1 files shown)


rust-query-crlite/src/main.rs
```diff
@@ -537,6 +537,7 @@ impl Intermediates {
                 if cert.verify_signature(Some(&issuer_spki)).is_ok() {
                     return Some(issuer_spki);
                 }
+                debug!("Not issued by {}", issuer.tbs_certificate.subject);
```
mozkeeler, line 540: Did you still want this debugging statement here?

mozkeeler (approved):

rust-query-crlite/src/main.rs
jschanck: eh, I guess it's not that useful. I removed it.

PR #302, opened 2024-06-11
rust-create-cascade: statsd telemetry

jschanck:
Let's push some telemetry to statsd so we can set up alerts through grafana.

The new probes are:
- `crlite.generate.revoked`: number of known revoked certificates
- `crlite.generate.not_revoked`: number of known not-revoked certificates
- `crlite.generate.filter_size`: filter size in bytes
- `crlite.generate.stash_size`: latest stash size in bytes
- `crlite.generate.revset_size`: uncompressed revocation set size in bytes
- `crlite.generate.time`: seconds spent in filter generation


mozkeeler (approved):
LGTM!

PR #303, opened 2024-06-13
ct-fetch: statsd telemetry

jschanck:
More statsd telemetry. We haven't been emitting the existing metrics because our production instance wasn't setting the statsd port environment variable. None of the data was particularly useful anyway.

New metrics:
- `ct-fetch.<instance>.<log>.coverage`: number of log entries we've seen divided by tree size.
- `ct-fetch.<instance>.sync.error`: negative health indicator.
- `ct-fetch.<instance>.sync.progress`: positive health indicator.
- `ct-fetch.<instance>.parse.error`: pinged when we find a seriously malformed certificate which prevents us from making further progress.
- `ct-fetch.<instance>.<log>.parse.error`: same as above, but keyed by log to aid diagnosis.


mozkeeler (approved):
SGTM!

PR #304, opened 2024-06-21
Plumb revocation reason codes through to the filter generation process

jschanck:
This PR changes the output of aggregate-crls such that each serial number is prefixed by a one byte revocation reason code. The data is still encoded as line-delimited ascii hex.

This code is currently running on the dev instance. I've added dashboard entries for the new statsd telemetry.

### DIFF
--- a/go/storage/localdiskbackend.go
+++ b/go/storage/localdiskbackend.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"bufio"
 	"context"
+	"encoding/hex"
 	"os"
 	"path/filepath"
 
@@ -36,6 +37,8 @@ func makeDirectoryIfNotExist(id string) error {
 	return nil
 }
 
+// Write a line delimited list of serial numbers to a text file. The serial
+// numbers are lowercase hex encoded.
 func (db *LocalDiskBackend) StoreKnownCertificateList(ctx context.Context, issuer types.Issuer,
 	serials []types.Serial) error {
 	path := filepath.Join(db.rootPath, issuer.ID())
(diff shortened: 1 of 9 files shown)


go/storage/localdiskbackend.go
```diff
@@ -69,3 +72,44 @@ func (db *LocalDiskBackend) StoreKnownCertificateList(ctx context.Context, issue
...
+	if err != nil {
+		return err
+	}
+	defer fd.Close()
+
+	writer := bufio.NewWriter(fd)
+	defer writer.Flush()
+
+	for _, s := range serials {
+		select {
```
mozkeeler, line 96: Is this really the right formatting for this? Or is this just a display issue because go uses tabs?

go/types.go
```diff
@@ -179,10 +189,51 @@ type TBSCertificateListWithRawSerials struct {
...
+	return reasonCode, nil
+}
+
+func (c RevokedCertificateWithRawSerial) SerialAndReason() (SerialAndReason, error) {
+	reason, err := c.Reason()
+	if err != nil {
+		return SerialAndReason{}, err
+	}
+
+	if reason < 0 || reason > 255 {
```
mozkeeler, line 229: Do we have a test for this?

mozkeeler (approved):
LGTM!

go/storage/localdiskbackend.go
jschanck: gofmt is happy, and it doesn't look weird on my screen... :shrug:

go/types.go
jschanck: Good catch. I should have had this in `Reason()` instead of `SerialAndReason()`. I added some tests.

go/storage/localdiskbackend.go
mozkeeler: I just meant that the body of the `select` doesn't increase indentation, but I guess that's the style?

jschanck: ah, yeah that's the style.

PR #305, opened 2024-07-12
Produce filters that encode a limited set of reason codes

jschanck:
With this PR you can pass `--reason-set all` to rust-create-cascade to produce a filter that encodes all revocations, or
`--reason-set specified` to produce a filter that encodes revocations with specified reason codes, or `--reason-set key-compromise` to produce a filter that encodes key compromise revocations.

I've updated crlite-generate.sh so that each kind of filter is produced and stored in cloud storage. I've not updated moz_kinto_publisher, so only the `--reason set all` filter will actually be published through remote settings.


mozkeeler (approved):
lgtm!

PR #306, opened 2024-07-16
Publish limited reason code filters to remote settings

jschanck:


rust-create-cascade/src/main.rs
```diff
@@ -172,7 +172,9 @@ impl RevokedSerialAndReasonIterator {
         match self.reason_set {
             ReasonSet::All => false,
             ReasonSet::Specified => *reason == Reason::Unspecified,
-            ReasonSet::KeyCompromise => *reason != Reason::KeyCompromise,
+            ReasonSet::Priority => {
+                !matches!(*reason, Reason::KeyCompromise | Reason::CessationOfOperation | Reason::PrivilegeWithdrawn)
```
mozkeeler, line 176: Is the reason we're not including CACompromise because if a CA were compromised, we would be revoking the appropriate intermediate/root?

mozkeeler (approved):
lgtm!

rust-create-cascade/src/main.rs
jschanck: Right, we would use OneCRL in that case. In fact, CRLs signed by intermediates in the Mozilla root program should not contain any CACompromise reasons: https://wiki.mozilla.org/CA/Revocation_Reasons#Banned_Revocation_Reasons

PR #307, opened 2024-07-18
Fix a few bugs discovered in staging testing of filter channels

jschanck:
Fixes three issues found in staging testing:
1) the publisher would fail before the first generate task finished because is_run_valid didn't check for existence of mlbf-specified and mlbf-priority files,
2) the publisher assumed that attachment filenames contain exactly two hyphens when checking run id consistency,
3) the signoff job would fail because rust-query-crlite didn't filter records by channel.

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -51,13 +51,23 @@ class PublishedRunDB:
     def __len__(self):
         return len(self.run_identifiers)
 
-    def is_run_valid(self, run_id):
+    def is_run_valid(self, run_id, channel):
+        if channel == CHANNEL_ALL:
+            mlbf_dir = "mlbf"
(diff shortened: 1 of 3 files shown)


moz_kinto_publisher/main.py
```diff
@@ -51,13 +51,23 @@ def __init__(self, filter_bucket):
     def __len__(self):
         return len(self.run_identifiers)
 
-    def is_run_valid(self, run_id):
+    def is_run_valid(self, run_id, channel):
```
mozkeeler, line 54: We also have this logic in `publish_crlite` - should this be a helper function?

mozkeeler (approved):
Makes sense!

moz_kinto_publisher/main.py
jschanck: Yep, done!

PR #308, opened 2024-07-19
moz_kinto_publisher: avoid overriding PublishedRunDB functions in MockRunDB tests

jschanck:
The MockRunDB class in test_publisher.py was overriding PublishedRunDB's `get_run_timestamps()`, `is_run_valid()`, and `is_run_ready()` methods, because those methods usually make network requests to google cloud storage. This has caused us to miss bugs in local testing that end up taking down our stage instance (like the typo in main.py that this PR fixes). The strategy here is to instead define a "bucket" that uses the local file system instead of GCS.

### DIFF
--- a/moz_kinto_publisher/test_publisher.py
+++ b/moz_kinto_publisher/test_publisher.py
@@ -2,6 +2,9 @@ import main
 import settings
 import unittest
 import hashlib
+import os
+import pytest
+import workflow
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
@@ -12,7 +15,7 @@ def timestamp_from_run_id(run_id):
     return datetime.strptime(time_string, "%Y%m%d-%H").replace(tzinfo=timezone.utc)
 
 
-def make_record(run_id, *, parent, channel):
+def make_record(run_id, *, parent, channel=main.CHANNEL_ALL):
     timestamp = timestamp_from_run_id(run_id)
     record_type = "diff" if parent else "full"
     record_time = timestamp.isoformat(timespec="seconds")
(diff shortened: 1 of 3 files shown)


moz_kinto_publisher/test_publisher.py
```diff
@@ -87,108 +94,113 @@ def test_load_remote(self):
 
 
 class TestPublishDecisions(unittest.TestCase):
+    # Create a new temporary directory and cd into it for each test.
```
mozkeeler, line 97: So are the temporary directories cleaned up automatically?

moz_kinto_publisher/test_publisher.py
```diff
@@ -45,17 +48,21 @@ def make_record(run_id, *, parent, channel):
...
-        return timestamp_from_run_id(run_id)
-
-    def is_run_valid(self, run_id, channel):
-        return run_id in self.run_identifiers
-
-    def is_run_ready(self, run_id):
-        return True
+    def __init__(self, runs, completed_runs, channel=main.CHANNEL_ALL):
+        for run in runs:
+            os.mkdir(Path(run))
```
mozkeeler, line 53: Do we need to register any kind of cleanup function for this?

mozkeeler (approved):
Makes sense. I'm just wondering if the test directories get cleaned up automatically?

moz_kinto_publisher/test_publisher.py
jschanck: Yes, everything here ends up living in pytest's "[base temporary directory](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-default-base-temporary-directory)".

moz_kinto_publisher/test_publisher.py
jschanck: Nope, we get a fresh temporary directory for each test.

PR #318, opened 2024-10-10
Add a new "experimental+deltas" filter channel

jschanck:
Start publishing clubcard-based delta updates on a new `experimental+deltas` channel. Also stop publishing the `specified` and `priority` channels.

### DIFF
--- a/containers/scripts/crlite-generate.sh
+++ b/containers/scripts/crlite-generate.sh
@@ -60,18 +60,6 @@ ${workflow}/2-generate_mlbf ${ID} \
 ${workflow}/2-generate_mlbf ${ID} \
               --filter-bucket ${crlite_filter_bucket:-crlite_filters_staging} \
               --statsd-host ${statsdHost} \
-              --reason-set specified \
-              --filter-type cascade
-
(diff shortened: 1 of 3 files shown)


containers/scripts/crlite-generate.sh
```diff
@@ -57,18 +57,6 @@ ${workflow}/2-generate_mlbf ${ID} \
               --reason-set all \
               --filter-type cascade
 
-${workflow}/2-generate_mlbf ${ID} \
-              --filter-bucket ${crlite_filter_bucket:-crlite_filters_staging} \
-              --statsd-host ${statsdHost} \
-              --reason-set specified \
```
mozkeeler, line 63: If we're not going to publish "specified" or "priority" filters, should we remove their mention from `modules/libpref/init/StaticPrefList.yaml`?

mozkeeler (approved):

containers/scripts/crlite-generate.sh
jschanck: Yeah, I guess I could do that in D225261. They'll still be available until we cut a new crlite release and push it to prod, but I plan to do that pretty quickly.

PR #321, opened 2024-10-22
moz_kinto_publisher: unpack noop argument

jschanck:


mozkeeler (approved):
Ah, sorry - I should have noticed this the first time.

PR #322, opened 2024-10-22
fixes #263: remove unnecessary intermediate attributes

### LINKED ISSUE #263: remove unnecessary attributes from intermediates collection
See https://bugzilla.mozilla.org/show_bug.cgi?id=1565282.
The `subject` attribute isn't used and is just taking up space, so it can be removed.
The `whitelist` attribute isn't used and uses problematic terminology, so it can be removed.

mozkeeler:
I think this may depend on schema changes in https://github.com/mozilla-services/remote-settings-permissions that I'll do shortly.


moz_kinto_publisher/main.py
```diff
@@ -373,15 +320,13 @@ def _upload_pem(self, *, rw_client=None, kinto_id=None):
         rw_client.attach_file(
             collection=settings.KINTO_INTERMEDIATES_COLLECTION,
             fileContents=self.pemData,
-            fileName=f"{base64.urlsafe_b64encode(self.pubKeyHash).decode('utf-8')}.pem",
+            fileName=f"{base64.urlsafe_b64encode(self.derHash).decode('utf-8')}.pem",
```
mozkeeler, line 323: I'm a bit concerned that this will update all existing records, but maybe that's what we want anyway? (to apply this change to current records)

jschanck (approved):
lgtm! I think we will need to wait for the schema changes.

moz_kinto_publisher/main.py
mozkeeler: Ok - I think it would be best to avoid that, so I reworked it to keep `unique_id()` how it was (and `fileName` here). How's this look?

PR #324, opened 2024-10-24
Keep an intermediate "enrolled" even when we fail to download a CRL

jschanck:
We would previously "unenroll" an issuer if we failed to download one of its CRLs for an extended period of time. The thinking being that we did not want to introduce false negatives into the system. When an issuer was unenrolled, we would not include any of its certificates in our filters. So filter coverage depended on both CT log timestamps and the list of enrolled issuers.

To simplify coverage checks, we're going to remove the enrollment bit. Every issuer in Mozilla's root program will be considered enrolled. We'll invest in our CRL monitoring processes to reduce the risk that this introduces false negatives.


mozkeeler (approved):
LGTM!

PR #325, opened 2024-10-25
rust-create-cascade: treat missing revoked cert file as an empty list when constructing clubcards

jschanck:
The clubcard library requires you to specify a subset of the chosen universe, even if it's the empty set. In rust-create-cascade were specifying the subset in `clubcard_do_one_issuer`, but we were only calling that function if the revoked certificate file existed. Due to #324, we're now more likely to have a known cert file that does not have a matching revoked cert file. So that PR broke the generate task for clubcards.

### DIFF
--- a/rust-create-cascade/src/clubcard_helper.rs
+++ b/rust-create-cascade/src/clubcard_helper.rs
@@ -69,7 +69,7 @@ impl FilterBuilder for ClubcardBuilder<4, CRLiteBuilderItem> {
                         Some(revoked_file) => {
                             RevokedSerialAndReasonIterator::new(revoked_file, reason_set)
                         }
-                        None => Default::default(), // empty iterator
+                        None => RevokedSerialAndReasonIterator::empty(reason_set),
                     };
                     clubcard_do_one_issuer(
                         self,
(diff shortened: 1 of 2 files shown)


rust-create-cascade/src/clubcard_helper.rs
```diff
@@ -62,21 +62,22 @@ impl FilterBuilder for ClubcardBuilder<4, CRLiteBuilderItem> {
...
-                        None
-                    }
+                .map(|(issuer, maybe_revoked_file, known_file)| {
+                    let issuer_bytes =
+                        decode_issuer(issuer.to_str().expect("non-unicode issuer string"));
+                    let revoked_serials_and_reasons = match maybe_revoked_file {
+                        Some(revoked_file) => {
+                            RevokedSerialAndReasonIterator::new(revoked_file, reason_set)
+                        }
+                        None => Default::default(), // empty iterator
```
mozkeeler, line 72: I'm a little concerned that the default `RevokedSerialAndReasonIterator` may have a different `reason_set` here, in terms of potential future changes. What do you think about passing `maybe_revoked_file` through to `RevokedSerialAndReasonIterator::new` instead of using `Default`?

mozkeeler (approved):
LGTM - just the one comment.

rust-create-cascade/src/clubcard_helper.rs
jschanck: That's reasonable. Changing `new` would require a bunch of changes in other parts of the code. So I added an explicit constructor for empty `RevokedSerialAndReasonIterator`s that takes a reason set.

mozkeeler: Ok!

jschanck:
Tacking on a patch to fix a test-only build failure.

PR #327, opened 2024-10-28
Plumb expiry data through to filter generation

jschanck:
This changes the format of "known" files. Previously they were unstructured lists of serial numbers. With this PR the serial numbers are split into blocks delimited by timestamp lines. The timestamp lines are hex-encoded 64 bit unix timestamps prefixed by an "@" character. A serial number in block `t` corresponds to a certificate with a notAfter date that truncates to `t` (the truncation is either to the hour or to the day depending on how the backend is configured).

This PR also adds statsd telemetry for the information theoretic lower bound for the size of a filter that is partitioned by  issuer+expiry. Ignoring telemetry, the expiry information is not used in filter generation.


rust-create-cascade/src/main.rs
```diff
@@ -227,20 +233,29 @@ impl From<RevokedSerialAndReasonIterator> for HashMap<Serial, Reason> {
...
-    type Item = String;
+    type Item = (u64, String);
     fn next(&mut self) -> Option<Self::Item> {
-        self.lines.next().transpose().expect("io error")
+        while let Some(line) = self.lines.next().transpose().expect("io error") {
+            if let Some(timestamp) = line.strip_prefix("@") {
+                self.date = u64::from_str_radix(timestamp, 16).expect("malformed date");
+                continue;
+            }
+            return Some((self.date, line));
```
mozkeeler, line 256: There should always be a timestamp first, so we should always have a value for `date`, right? Maybe this should be an `Option` and we assert that it's `Some`?

rust-create-cascade/src/main.rs
```diff
@@ -823,12 +888,21 @@ fn main() {
...
-    let (delta_revoked, delta_not_revoked, delta_reasons, delta_lower_bound) =
-        count_all(delta_dir, known_dir, delta_reason_set, None);
-    info!("Lower bound is {:.0} bytes", delta_lower_bound / 8.0);
+    let delta_stats = count_all(delta_dir, known_dir, delta_reason_set, None);
+    info!(
+        "Lower bound is {:.0} bytes",
+        delta_stats.split_by_issuer_lower_bound
+    );
+    info!(
+        "Lower bound is {:.0} bytes",
```
mozkeeler, line 897: "Lower bound by issuer and serial number"?

mozkeeler (approved):
Sounds good! Just a couple of comments.

rust-create-cascade/src/main.rs
jschanck: I have old data sets that I use for testing, so I wanted this to work even if there were no "@timestamp" lines.

mozkeeler: Ok!

PR #328, opened 2024-10-28
ct-fetch: fix nil pointer dereference

jschanck:


go/cmd/ct-fetch/ct-fetch.go
```diff
@@ -504,7 +504,9 @@ func (ld *LogSyncEngine) NewLogWorker(ctx context.Context, ctLogMeta *types.CTLo
 	}
 
 	metricKey := ctLogMeta.MetricKey()
-	metrics.SetGauge([]string{metricKey, "coverage"}, float32(logObj.MaxEntry-logObj.MinEntry+1)/float32(sth.TreeSize))
+	if sth != nil {
```
mozkeeler, line 507: Don't we already dereference `sth` on e.g. line 478?

mozkeeler: Ohh, right, not if `fetchErr` is something...

jschanck: Right, `sth` is only `nil` if `fetchErr` is not.

mozkeeler (approved):
LGTM. It seems like in other places we use `fetchErr == nil` as the "is-sth-valid" guard, but I'm not sure that's better.

PR #330, opened 2024-11-01
moz_kinto_publisher: extract subject and spki hash from the certificate when they are not in the remote settings record

jschanck:


moz_kinto_publisher/main.py
```diff
@@ -376,14 +381,28 @@ def set_pem(self, pem_data):
...
             raise IntermediateRecordError("Cannot parse PEM data: {}".format(e))
 
+        derHash = hashlib.sha256(self.cert.public_bytes(Encoding.DER)).digest()
+        if self.derHash and self.derHash != derHash:
+            raise IntermediateRecordError("DER hash does not match")
+        self.derHash = derHash
+
+        self.subject = self.cert.subject.rfc4514_string()
+
+        der_spki = self.cert.public_key().public_bytes(
```
mozkeeler, line 398: nit: it seems like the style in this file would be `derSpki`, but it's not a big deal

mozkeeler (approved):
LGTM - thanks for taking care of this!

PR #331, opened 2024-11-07
Rework channels in moz_kinto_publisher

jschanck:
This is a grab-bag of changes to the channels published by moz_kinto_publisher.

1) Remove the specified and priority channels entirely---all of the records associated with these channels have been removed, so there's no need to keep the definitions around.
2) Disable the `experimental` channel. This was only ever used by Nightly 133 users. It has been superseded by the experimental+deltas channel, which is also supported by 133.
3) Replace the 10-day filter age cutoff with a tunable parameter, and set it to 20 days for the experimental+deltas channel.
4) Allow the total size of delta files (but not stashes) to exceed the size of the full filter.

### DIFF
--- a/moz_kinto_publisher/main.py
+++ b/moz_kinto_publisher/main.py
@@ -66,7 +66,7 @@ CHANNELS = [
         delta_filename="filter.stash",
         supported_version=132,
         mimetype="application/octet-stream",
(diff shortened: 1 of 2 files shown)


mozkeeler (approved):
LGTM!

PR #332, opened 2025-01-07
Remove unused code related to cache and storage types

jschanck:
The biggest change here is the removal of `IssuerMetadata`. We stopped using the CRL DP and Issuer DN data tracked by `IssuerMetadata` some time ago, but we were still saving that data in Redis.


go/storage/serialcachewriter.go
```diff
@@ -10,33 +10,37 @@ import (
...
+	return &SerialCacheWriter{
 		expDate:   aExpDate,
 		issuer:    aIssuer,
 		cache:     aCache,
 		expirySet: false,
 	}
 }
 
-func (kc *KnownCertificates) id(params ...string) string {
+func NewSerialCacheReader(aExpDate types.ExpDate, aIssuer types.Issuer, aCache RemoteCache) *SerialCacheWriter {
```
mozkeeler, line 29: I do find it a bit odd that a function called `NewSerialCacheReader` returns a `SerialCacheWriter`, but I guess I understand the intention.

mozkeeler (approved):
LGTM!

go/storage/serialcachewriter.go
jschanck: The sole caller of this function is going to be removed in a follow-up commit. I did this as a note-to-self that the new code should have separate reader and writer types.

PR #334, opened 2025-01-23
Stop publishing the cascade-based "all" channel and add the "default" channel

jschanck:
The default channel for Firefox users is going to be clubcard-based CRLite with 12 hour latency and a 45 day recompression interval. This patch adds that channel. It also removes the "all" channel, which was still based on Bloom filter cascades.

I've taken the opportunity to remove support for cascades from moz_kinto_publisher and rust-query-crlite. And I've also added support for delta clubcards to rust-query-crlite.


mozkeeler (approved):
LGTM!

PR #340, opened 2025-02-12
storage: limit number of goroutines used in moveCachedSerialsToStorage

jschanck:
It turns out that one issuer in our database has certificates with ~17000 distinct expiry-hour values (about 2 years worth if they're sequential). We hit a `runtime: program exceeds 10000-thread limit` error while processing that issuer. It's not clear to me why the go scheduler doesn't handle this for us..

This PR makes it so we process no more than 1000 expiry-hour bins at a time.

### DIFF
--- a/go/storage/certdatabase.go
+++ b/go/storage/certdatabase.go
@@ -23,7 +23,8 @@ import (
 )
 
 const (
-	permModeDir = 0755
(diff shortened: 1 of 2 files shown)


mozkeeler (approved):
LGTM

PR #345, opened 2025-03-13
Revert "aggregate-known: log unknown issuers"

jschanck:
This is actually very noisy and not as helpful as I'd hoped.


mozkeeler (approved):
Ok!

PR #348, opened 2025-04-14
moz_kinto_publisher: enroll new CT logs in CRLite monitoring by default

jschanck:
This will cause us to automatically start monitoring CT logs when they appear in Google's collection. We were previously enrolling logs manually, mostly due to concerns about running up against our Redis memory limit. This is no longer a concern after #333.


mozkeeler (approved):
Great!

PR #357, opened 2025-07-01
Validate issuing dp

### LINKED ISSUE #356: Check CRL Issuing Distribution Point extension of partitioned CRLs
From Corey Bonnell via security@mozilla.org:
> It appears that there is no logic to compare CRL URLs against URLs contained within the CRL Issuing Distribution Point extension. This check is needed to ensure that a substitution attack has not occurred where an attacker has tampered with the HTTP connection and served a different CRL issued by the same CA. The security implications of not performing this check have been discussed on MDSP and in other venues over the years, with the most recent here [2].
> [...]
> [2] https://groups.google.com/a/mozilla.org/g/dev-security-policy/c/qhrGxLvyreU

For each URL that is listed in a "JSON Array of Partitioned CRLs" field of the CCADB report, we should ensure that the CRL that we receive from that URL lists that URL in its Issuing Distribution Point extension. Failure to perform this check lets an on-path attacker reduce the CRLite aggregator's view to a subset of an issuer's CRLs.

jschanck:
Add a new CRL audit entry type for when a Partitioned CRL's `issuingDistributionPoint` extension does not match the URL that the CRL was fetched from.

Resolves #356.

### DIFF
--- a/go/cmd/aggregate-crls/aggregate-crls.go
+++ b/go/cmd/aggregate-crls/aggregate-crls.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"encoding/pem"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


go/rootprogram/issuers.go
```diff
@@ -233,7 +234,7 @@ func (mi *MozIssuers) LoadEnrolledIssuers(filePath string) error {
 		if err != nil {
 			return err
 		}
-		mi.InsertIssuerFromCertAndPem(cert, ei.Pem, nil)
+		mi.InsertIssuerFromCertAndPem(cert, ei.Pem, nil, false)
```
mozkeeler, line 237: Does go have a maybe/option type? It would be nice to ensure we're not accidentally using the value of `usesPartitionedCrls` when we don't load from the ccadb.

go/cmd/aggregate-crls/aggregate-crls_test.go
```diff
@@ -659,3 +660,74 @@ func Test_SaveRevokedCertificateList(t *testing.T) {
...
+MTE2MDIwNVoXDTI1MDcxMTE1MDIwNFowAKAuMCwwKgYDVR0cAQH/BCAwHqAcoBqG
+GGh0dHA6Ly9leGFtcGxlLmNvbS8xLmNybDAKBggqhkjOPQQDAgNHADBEAiAAAAAA
+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIgAAAAAAAAAAAAAAAAAAAAAAAA
+AAAAAAAAAAAAAAAAAAA=
+-----END X509 CRL-----`))
+
+	crl, _ := x509.ParseCertificateListDER(crlBlock.Bytes)
+
+	fetchUrl := "http://example.com/1.crl"
+	if !crlIssuingDPMatches(crl, fetchUrl) {
```
mozkeeler, line 725: Am I missing something? I'm not seeing where this is defined... (is the review UI messing me up?)

mozkeeler (approved):
LGTM! Just a couple of comments.

go/rootprogram/issuers.go
jschanck: I updated this so now we pipe the `usesPartitionedCrls` flag through the enrolled.json file for use here.

go/cmd/aggregate-crls/aggregate-crls_test.go
jschanck: Whoops, refactored and forgot to update this test.

go/cmd/aggregate-crls/aggregate-crls.go
```diff
@@ -188,6 +189,49 @@ func loadAndCheckSignatureOfCRL(aPath string, aIssuerCert *x509.Certificate) (*p
...
+	if err != nil {
+		return false, fmt.Errorf("Error parsing, will not process revocations: %s", err)
+	}
+
+	urls := []string{}
+	found := false
+	for _, url := range crl.TBSCertList.IssuingDPFullNames.URIs {
+		urls = append(urls, url)
+		_, exists := aPartitionedCrlUrlSet[url]
+		if exists {
```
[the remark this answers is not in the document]
jschanck: Right, it's ok for there to be multiple URLs for the same CRL in `issuingDistributionPoints`. But only one URL from `issuingDistributionPoints` should appear in `aPartitionedCrlUrlSet` since the URLs in that set must point to distinct CRLs.

PR #361, opened 2025-07-21
ct-fetch: use one insertCTWorker per log

jschanck:
In #359 I switched to using a single `insertCTWorker`. I thought this would not negatively impact performance because Redis is single threaded. However, each operation with the go-redis client incurs a round-trip network time delay. The previous parallel strategy, while unsound, was reducing the time between database writes.

The right fix here is to pipeline our go-redis client operations, so that a batch of database writes are sent in the same network payload. In the meantime, we can use one `insertCTWorker` per log. This will increase throughput while still ensuring that writes from a single LogWorker are sequenced correctly.


jschanck:
On second thought I'll just implement the pipelining.

PR #362, opened 2025-07-23
ct-fetch: use new sunlight.Client.Fetcher().ReadEndpoint

FiloSottile:
I noticed your client from the User-Agent logs, and thought it was silly to make you reimplement ReadEndpoint, so I exposed it on the Client.


jschanck (approved):
Awesome, thank you!

PR #369, opened 2025-09-24
storage: hold serials in cache until they are covered

### LINKED ISSUE #367: Certificates revoked within one CT log MMD of issuance can fail to appear as revoked in delta updates
Mozilla [Bug 1988663](https://bugzilla.mozilla.org/show_bug.cgi?id=1988663) reports that [the certificate](https://crt.sh/?id=20924740030) sent by https://revoked-isrgrootx1.letsencrypt.org/ is not marked as revoked in the current set of CRLite filters. I've determined that this is because it was not covered by CT log metadata in the first filter in which it was known and marked as revoked.

**Upshot**: Certificates that are revoked within the merge delay of the log in which they were discovered may not be marked as revoked until the next full snapshot filter is published.

**Explanation**: Suppose we scrape a CT log that has an MMD of `δ` seconds. If a certificate is discovered in that log at time `t_0` and filter generation runs at time `t_1` with `t_0 < t_1 < t_0 + δ`, then that certificate will have the `NotCovered` status in the first clubcard that includes it. If the certificate appears in a CRL at time `t_1`, then the certificate will be in the "known revoked" set at all times `> t_1`. Hence, the certificate will not be included in a delta updates generated at time `t > t_1` and will have the `Good` status in those filters.

The certificate will be marked as "revoked" in any full snapshot filter generated at time `t > t_0 + δ` (so long as the backend has ingested an SCT from a log with a timestamp that is greater than `t_0 + δ`). But in our current configuration this could be up to 45 days after issuance.

**Remediation**: I see two options here.
1) Wait to add a certificate to the known set until it is covered. This is difficult to do with our current pipeline because we only store the certificate's serial number, issuer SPKI hash, and expiry hour. To evaluate coverage, we would need an SCT as well (or at least a log ID and timestamp).
2) Do nothing. The problem goes away if `δ = 0` and static CT logs have very short MMDs (if not `δ = 0`, then something like `δ = 60` seconds). CAs are going to begin production logging to static CT logs in the immediate future, possibly even before we could ship option 1.

Certificates that are revoked within 1 MMD of issuance are unlikely to have been revoked for a security-relevant reason. So while it is unfortunate that highly-visible CA test sites like https://revoked-isrgrootx1.letsencrypt.org/ are affected, the security impact of this issue is small.

jschanck:
This changes the redis cache so that it holds tuples (timestamp, log id, serial) rather than raw serials. It also changes the Commit process so that cache entries are only moved to storage if the MaxTimestamp for the corresponding log is at least one MMD ahead of the timestamp value for the entry.

One minor hangup is that the cache will hold serial numbers in the legacy format when we roll this out. The baseline requirements say that serial numbers cannot be more than 20 bytes. So we can identify legacy cache entries by their length (timestamp is 8 bytes and log id is 32, so new cache entries are at least 40 bytes).

There could be a legacy cache entry that is in violation of the BRs and is more than 40 bytes. Such an entry will persist in cache until the corresponding certificate expires. There will be a risk that CRLite filters give it the wrong status. The risk here is quite low as such a certificate would need to be issued in the 12 hour period when we roll this out. And also the BRs would require that certificate to be revoked anyway.

Resolves #367

### DIFF
--- a/go/cmd/aggregate-known/aggregate-known.go
+++ b/go/cmd/aggregate-known/aggregate-known.go
@@ -86,16 +86,8 @@ func (kw knownWorker) run(ctx context.Context, wg *sync.WaitGroup, workChan <-ch
 				if err != nil {
 					glog.Fatalf("[%s] Could not read serials with expDate=%s: %s", tuple.issuer.ID(), expDate.ID(), err)
 				}
-				knownSetLen := uint64(len(knownSet))
 
-				if knownSetLen == 0 {
-					// This is almost certainly due to an hour-rollover since the loader ran, and expired all the next hour's
-					// certs.
-					glog.Warningf("No cached certificates for issuer=%s (%s) expDate=%s, but the loader thought there should be."+
-						" (current count this worker=%d)", tuple.issuerDN, tuple.issuer.ID(), expDate, serialCount)
-				}
-
-				serialCount += knownSetLen
+				serialCount += uint64(len(knownSet))
 				err = storage.WriteSerialList(writer, expDate, tuple.issuer, knownSet)
 				if err != nil {
 					glog.Fatalf("[%s] Could not write serials: %s", tuple.issuer.ID(), err)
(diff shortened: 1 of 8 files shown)


mozkeeler (approved):
LGTM!

PR #371, opened 2025-09-25
storage: remove preissuer alias handling

jschanck:
The mitigation for Bug 1955023 introduced in #346 has been in place for 6 months. We stopped adding new aliases in #359, so all of the aliased bins in our production instance are empty. We can remove the code for moving aliased serials to storage now.


mozkeeler (approved):
Great!

PR #373, opened 2025-10-02
add a standalone program to commit the redis cache to disk

### LINKED ISSUE #370: Commit cache to storage more frequently
The `aggregate-known` program moves serials from cache to persistent storage. We currently run `aggregate-known` in the [crlite-generate.sh](https://github.com/mozilla/crlite/blob/main/containers/scripts/crlite-generate.sh) script, which only runs once every 12 hours in our production environment.

It would be nice to move serials from cache to storage more frequently. This would reduce memory pressure on the cache and reduce the risk that we pause log ingestion while we wait for memory to become available.

jschanck:
Resolves #370 

We'll also need to add a cronjob to run crlite-commit.sh hourly, but that lives in a private cloudops repo.


mozkeeler (approved):
SGTM!

PR #374, opened 2025-10-02
aggregate-known: wait up to 15 minutes for commit lock

jschanck:
The aggregate-known job only runs once every 12 hours, and it would be a shame to miss a chance to run just because another process was committing the redis cache to disk. This adds a retry loop for trying to acquire the commit lock. It will retry once per second for up to 15 minutes.


mozkeeler (approved):
Ok!

PR #383, opened 2026-02-18
rust-query-crlite: use enrolled.json rather than ccadb report

jschanck:
The filter generation process produces a nice JSON file from the CCADB CSV report that the backend uses. Rather than worry about parsing the CSV correctly, we can have rust-query-crlite use the latest enrolled.json file.

### DIFF
--- a/rust-query-crlite/src/main.rs
+++ b/rust-query-crlite/src/main.rs
@@ -102,17 +102,13 @@ struct EnrolledRecord {
     pem: String,
 }
 
-fn extract_filter_prefix(path: &Path) -> Option<String> {
+fn extract_run_id(path: &Path) -> Option<(String, u32)> {
     let filename = path.file_name()?.to_str()?;
     let mut parts = filename.splitn(3, '-');
     let date = parts.next()?;
-    let revision = parts.next()?;
-    if date.len() == 8
-        && date.chars().all(|c| c.is_ascii_digit())
-        && !revision.is_empty()
-        && revision.chars().all(|c| c.is_ascii_digit())
-    {
-        Some(format!("{}-{}", date, revision))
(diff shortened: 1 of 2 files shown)


rust-query-crlite/src/main.rs
```diff
@@ -95,20 +95,57 @@ struct CertRevRecordAttachment {
...
+    }
+}
 
-    let intermediates = Intermediates::from_ccadb_csv(intermediates_bytes)
-        .map_err(|_| CRLiteDBError::from("cannot parse CCADB report"))?;
+fn update_intermediates(db_dir: &Path) -> Result<(), CRLiteDBError> {
+    let intermediates_path = db_dir.join("crlite.intermediates");
 
-    let encoded_intermediates = intermediates.encode()?;
+    // Find the lexicographically largest "yyyymmdd-r" prefix among downloaded filter files.
```
mozkeeler, line 120: Is this going to do the right thing if we ever publish more than 9 revisions in a day? (e.g., "20260218-10" I believe will sort before "20260218-9", but we'd want the former here, right?)

mozkeeler (approved):
LGTM - just the one comment.

rust-query-crlite/src/main.rs
jschanck: Yeah, although we shouldn't have 10+ revisions and any recent enrolled.json is good enough. There are corner cases regardless of which enrolled.json you download. For example, if an intermediate drops out of the collection between revisions 1 and 2 you can get a false `NotEnrolled` result when querying the revision 1 filter. All that would be fixed if this program took both an intermediate and an end-entity as input, but I like the convenience of being able to provide just an end-entity.

PR #385, opened 2026-05-11
crlite-signoff: stop enforcing verification_domains.txt check.

jschanck:
The verification domains check is supposed to stop us from publishing a filter that would block access to critical infrastructure. It fetches certificates from a list of TLS servers and checks that none of them are marked as revoked. The check fails closed, so a temporary outage at one of the servers or a failure to fetch, say, the CCADB intermediate certificate list causes us to halt filter publication.

The verification domains check was plausibly useful when our CRLite implementation suffered from false positives. Now it just has a tendency to halt publication. If we were to ever publish a filter that blocked access to critical Firefox infrastructure, we would simply rotate the certificate on the affected server and/or publish new filters.


mozkeeler (approved):
Sounds good to me.

PR #386, opened 2026-05-21
signoff: ping statsd crlite.signoff.cert_revocations when a new filter is published

jschanck:
This will allow us to set up alerting through Grafana. If there's no ping within, say, 14 hours we'll fire off an alert.


mozkeeler (approved):
Ok!

PR #388, opened 2026-07-31
Save v4 encoded clubcards in cloud storage

jschanck:
We currently create, store, and publish Clubcards in the v3 (bincode) format. As a first step towards switching to the v4 format, this PR saves v4 encoded clubcards in our public cloud storage. This will allow non-Firefox consumers of CRLite data to switch to the v4 format before Firefox.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -120,7 +120,7 @@ jobs:
           command: rustc --version; cargo --version; rustup --version
       - run:
           name: rustfmt
-          command: rustfmt --check src/*.rs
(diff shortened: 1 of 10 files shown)


mozkeeler (approved):

jschanck:
This is live now.

Example URLs:
https://storage.googleapis.com/crlite-filters-prod/20260807-1/clubcard-all/filter.v4
https://storage.googleapis.com/crlite-filters-prod/20260807-1/clubcard-all/filter.delta.v4

PR #389, opened 2026-07-31
rust-create-cascade: invert blocks where more than 2/3 of known certs are revoked

jschanck:
(Stacked on top of #388)

The clubcard crate has long supported "inverted blocks" which encode the complement of a set rather than the set itself. Prior to this PR we were only using an inverted block when all of the known certificates for a CA were revoked. This PR uses an inverted block whenever doing so drops the expected size of a filter.

For the 2026-07-27-0 data set, this increases the number of inverted blocks from 19 to 47 and it decreases the filter size by 24 kB. Much larger savings can be expected during mass revocation events.

Encoding an inverted block requires an extra pass over the set of known certificates. This is, at worst, a factor 2 increase in runtime. In practice,  the runtime of filter generation is dominated by a few large CAs with small revocation rates, and the runtime increase from handling the few small CAs with high revocation rates is negligible.

### DIFF
--- a/rust-create-cascade/src/clubcard_helper.rs
+++ b/rust-create-cascade/src/clubcard_helper.rs
@@ -20,26 +20,93 @@ use std::io::prelude::Write;
 use std::io::BufReader;
 use std::path::Path;
 
+/// `estimated_block_bits` estimates the size of the pair of ribbons that encodes a
+/// `subset_size` element subset of a `universe_size` element universe.
+///
+/// This mirrors the sizing rules in clubcard's builder: the approximate ribbon is `rank`
+/// columns of (1+epsilon)*subset_size rows, and the exact ribbon is one column with a row
+/// for each item that survives the approximate filter, i.e. the subset itself plus a
+/// 2^-rank fraction of its complement.
+fn estimated_block_bits(subset_size: usize, universe_size: usize) -> f64 {
+    const EPSILON: f64 = 0.02;
+    if subset_size == 0 || universe_size == 0 {
+        return 0.0;
+    }
+    let subset = subset_size as f64;
+    let universe = universe_size as f64;
+    let rank = if 2 * subset_size >= universe_size {
+        0.0
+    } else {
+        ((universe - subset) / subset).log2().floor()
+    };
+    let survivors = subset + (universe - subset) * (-rank).exp2();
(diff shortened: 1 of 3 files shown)


jschanck:
Depends on https://github.com/mozilla/clubcard/pull/8

rust-create-cascade/src/clubcard_helper.rs
```diff
@@ -20,26 +20,93 @@ use std::io::prelude::Write;
...
+        );
+        // We buffered the revoked serials rather than the (larger) known set, so we have to
+        // re-read the known file to enumerate its complement. This only happens for the few
+        // issuers that we invert. Note that a serial listed twice in the known file yields a
+        // duplicate equation, which is redundant rather than exceptional, so there's no need
+        // to deduplicate the non-revoked serials here.
+        let revoked_set: HashSet<Serial> = revoked_in_universe.into_iter().collect();
+        for (_expiry, serial) in KnownSerialIterator::new(known_file) {
+            if !revoked_set.contains(&serial) {
+                ribbon_builder.insert(CRLiteBuilderItem::not_revoked(
```
mozkeeler, line 102: Just to confirm, saying `not_revoked` here and also setting `inverted` on the ribbon builder above doesn't result in some sort of double-negative situation?

rust-create-cascade/src/main.rs
```diff
@@ -1383,4 +1383,33 @@ mod tests {
...
+            &env.known_dir(),
+            &env.ct_logs_path(),
+            ReasonSet::All,
+            Encoding::V4,
+        );
+
+        let clubcard =
+            CRLiteClubcard::from_bytes(&clubcard_bytes).expect("cannot deserialize clubcard");
+        let issuer_bytes = decode_issuer(&issuer);
+        assert!(clubcard.as_ref().index()[issuer_bytes.as_slice()].inverted);
```
mozkeeler, line 1414: Should we also check that e.g. the clubcard in the previous test isn't inverted?

mozkeeler (approved):
LGTM - just a couple of comments.

rust-create-cascade/src/clubcard_helper.rs
jschanck: This is the correct usage, but the API could be safer. I'll add an assertion as part of https://github.com/mozilla/clubcard/pull/8/.

